PR #39, opened 2019-06-05
fix(signal/scan.py): final timestep

hemmelig:
If the timespan for detect was not an integer multiple the timstep, the final
timestep was being missed off as the endtime was seen as a strict limit
(in a while st + x < et loop). To rectify this problem, we now take the
integer ceiling of (et - st) / time_step and for loop this many times.


hemmelig:
Care must be taken, if this is being used, to properly merge the .scnmseed files after running all detect stages.

PR #69, opened 2020-06-03
chore(templates): Add Issue/Pull Request templates

hemmelig:
Add simple templates for new Issues or Pull Requests. There are two options
for Issues - Bug Reports and Feature Requests. More may be added in future.


CONTRIBUTORS.md
```diff
@@ -0,0 +1,6 @@
+* Julian Drew
```
twinder36, line 1: Alphabetical ordering? (not sure if that is the convention, but is what I remember seeing in most places)

PR #73, opened 2020-06-29
Rewrite setup.py and add docs

hemmelig:
This PR covers the addition of documentation to be hosted on ReadTheDocs. However, before this could be done, it was necessary to rewrite setup.py to build the C library using the setuptools/distutils toolkits and allow for this stage to be skipped when building the package on the RTD servers. This has allowed us to progress towards platform independence by providing platform specific compiler instructions. This installation process has been successfully tested against Ubuntu (16.04, 18.04) and MacOS. It is anticipated that the installation process should work on Windows, but we provide no guarantee at this stage.

This PR will introduce a Breaking Change - users will have to recompile their version of QuakeMigrate (by re-installing with `setup.py`). If the code was previously setup in 'install' mode, the user will have to manually remove the build library in the site-packages directory of their virtual environment.

The important changes include:

- Rewrote setup.py to build the linked library of C functions. Added a small module (`QMigrate.core.libnames.py`) to handle loading the C library code in a platform/compiler independent manner e16104e
- Added documentation to the C functions c0a9c58
- Added configuration for ReadTheDocs and documentation f980cad 0c52e15

In addition, some minor changes:

- Minor docstring updates (df53fec) 
- An internal change to the name of the Python module that provides the bindings to the compiled C library - `QMigrate.core.QMigratelib.py` -> `QMigrate.core.lib.py`


QMigrate/core/libnames.py
```diff
@@ -0,0 +1,74 @@
...
+    """
+
+    # our custom defined part of the extension file name
+    libdir = pathlib.Path(__file__).parent / "src"
+    libpath = _get_lib_path(libdir, name)
+    try:
+        cdll = ctypes.CDLL(libpath)
+    except Exception as e:
+        msg = f"Could not load shared library '{libpath.name}'.\n\n {e}"
+        raise ImportError(msg)
```
hemmelig, line 72:
```suggestion
    except Exception as e:
        print(f"Could not load shared library '{libpath.name}'.\n {e}"
              "Falling back on old dynamic library name.\n"
              "Consider recompiling the C library with setup.py.\n\n")
        cdll = ctypes.CDLL(libdir  / "QMigrate.so")
```

This would (hopefully) provide a soft deprecation for current users.

requirements.txt
```diff
@@ -0,0 +1,7 @@
+matplotlib
+numpy
+obspy
+pandas
+pyproj
+scipy
+scikit-fmm==2019.1.30
```
twinder36, line 7: Are our requirements actually more specific? (If so, is it helpful to list those specifics here?)

hemmelig: Currently, no? If a future dependency breaks, it’s better not to be pinned to a specific version. This way we can address the issue (fix or pin) as it comes up. Will depend on what you find with obspy inventory, but I think this is fine.

.readthedocs.yml
```diff
@@ -0,0 +1,28 @@
...
+# Build documentation with MkDocs
+#mkdocs:
+#  configuration: mkdocs.yml
+
+# Optionally build your docs in additional formats such as PDF and ePub
+formats: all
+
+# Optionally set the version of Python and requirements required to build your docs
+python:
+  version: 3.7
```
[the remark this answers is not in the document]
hemmelig: Just specifies a version for the read the docs server to use to install package and build docs.

docs/references.rst
```diff
@@ -0,0 +1,52 @@
...
+.. _Poiata-2016:
+
+    Poiata, N., Satriano, C., Vilotte, J. P., Bernard, P., & Obara, K. (2016). Multiband array detection and location of seismic sources recorded by dense seismic networks. Geophysical Journal International, 205(3), 1548-1573.
+
+.. _Rawlinson-2004:
+
+    Rawlinson, N. & Sambridge, M. (2004). 'Wavefront evolution in strongly heterogeneous layered media using the fast marching method'. Geophysical Journal International, vol 156, pp. 631-647. Rawlinson, N. & Sambridge, M. (2005). 'The fast marching method: An effective tool for tomographic imaging and tracking multiple phases in complex layered media'. Exploration Geophysics, vol 36, pp. 341-350.
+
+.. _Smith-2020:
+
```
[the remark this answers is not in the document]
hemmelig: Yeah - I think I dumped these in from the earlier paper draft - perhaps remove this from the commit for now and we can add back in once finalised?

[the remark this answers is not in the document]
twinder36: Sounds like a plan

requirements.txt
twinder36: I mean in terms of minimum versions that we require. So pandas > 1.0.0 etc. For obspy it would be similar (haven't checked yet), something like obspy > 1.2.0.

Again, not certain that there actually are any other cases of this.

setup.py
```diff
@@ -89,48 +68,161 @@ def find_meta(meta):
...
+        "author_email": find_meta("email"),
+        "license": find_meta("license"),
+        "classifiers": [
+            "Development Status :: Beta",
+            "Intended Audience :: Science/Research",
+            "Topic :: Scientific/Engineering",
+            "Natural Language :: English",
+            "License :: OSI Approved :: MIT License",
+            "Operating System :: OS Independent",
+            "Programming Language :: Python :: 3.6",
```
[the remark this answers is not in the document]
hemmelig: Need to add 3.8

requirements.txt
twinder36: I'm sure there is some level of requirement for pyproj, scipy, numpy and matplotlib. Whether it's worth digging into, or safe enough to rely on the obspy requirements to sort us out I don't know. They have:
`INSTALL_REQUIRES = [
    'numpy>=1.15.0',
    'scipy>=1.0.0',
    'matplotlib>=3.2.0',
    'lxml',
    'setuptools',
    'sqlalchemy',
    'decorator',
    'requests']`

I had a look back to when I had trouble with pyproj before, can't see any obvious breaking point, so I would leave that unless we have issues.

docs/tutorials/lut.rst
```diff
@@ -59,35 +59,35 @@ In all cases we will make use of the :mod:`QMigrate.io` module, so let's import
...
+    stations = read_stations("/path/to/station_file")
 
 Homogeneous velocity model
 ##########################
 Simply calculates the straight line traveltimes between stations and points in the grid.
 
 ::
 
-	qlut.compute(lut, stations, method="homogeneous", vp=5000., vs=3000.)
+	create_lut.compute(lut, stations, method="homogeneous", vp=5000., vs=3000.)
```
twinder36, line 72: In the other example scripts it's currently:

`from QMigrate.lut import compute, LUT`

`compute(lut, stations, method="homogeneous", vp=vp, vs=vs)`

I haven't thought about which is better, but should be consistent

docs/tutorials/lut.rst
```diff
@@ -59,35 +59,35 @@ In all cases we will make use of the :mod:`QMigrate.io` module, so let's import
...
 Fast-marching method
 ####################
 The fast-marching method implicitly tracks the evolution of the wavefront. See Rawlinson & Sambridge (2005) for more details.
 
 ::
 
-	vmod = qio.read_vmodel("/path/to/vmodel_file")
-	qlut.compute(lut, stations, method="1dfmm", vmod=vmod)
+	vmod = read_vmodel("/path/to/vmodel_file")
+	create_lut.compute(lut, stations, method="1dfmm", vmod=vmod)
```
twinder36, line 81: Ditto

docs/tutorials/lut.rst
```diff
@@ -59,35 +59,35 @@ In all cases we will make use of the :mod:`QMigrate.io` module, so let's import
...
 NonLinLoc style 2-D sweep
 #########################
 Uses the Eikonal solver from NonLinLoc under the hood to generate a traveltime grid for the 2-D slice that passes through the station and the point in the grid furthest away from that station. This slice is then "swept" using a bilinear interpolation scheme to produce a 3-D traveltime grid. This has the benefit of being able to include stations outside of the volume of interest, without having to increase the size of the grid.
 
 ::
 
-	vmod = qio.read_vmodel("/path/to/vmodel_file")
-	qlut.compute(lut, stations, method="1dsweep", vmod=vmod, block_model=True)
+	vmod = read_vmodel("/path/to/vmodel_file")
+	create_lut.compute(lut, stations, method="1dsweep", vmod=vmod, block_model=True)
```
twinder36, line 90: Ditto

docs/tutorials/lut.rst
```diff
@@ -59,35 +59,35 @@ In all cases we will make use of the :mod:`QMigrate.io` module, so let's import
 
 ::
 
-    import QMigrate.io as qio
+    from QMigrate.io import read_stations, read_vmodel
```
twinder36, line 62: Hasn't been changed to read_stations yet? But if you're going to change this imminently then all good. (But note it needs changing in the other examples.)

hemmelig: Pushed this up to refactor_magnitudes (e89c3b0)

hemmelig:
> Only remaining comments:
> 
> * Do we want to add something in the install about using system gcc vs. conda gcc for Linux? Or just stick with the conda instruction as it's (hopefully) safer? If system gcc works on the vast majority of OS's it would be easier..
> 
> * Do you want to add the STEIM1 encoding option to `ScanmSEED.write()`

I think we are safe to fall back on system compilers and include instructions on installing a compiler with Conda if the user doesn’t have one.

Yes, will add now

PR #76, opened 2020-07-05
[WIP] Refactor event, output files and export.to_obspy

twinder36:
This PR aims to tidy and internalise the handling of event attributes, review the .event file output, and set floating point precision here and in the .picks and .amps outputs. Scripts in the `QMigrate.io.export` submodule will then be modified accordingly.

Some of these changes are required to be made in tandem with changing / adding attributes to the LUT module, which is also under review. As such, some issues raised here will be have to be resolved in coordination with those changes.

**Aims**
- [x] Internalise event dataframe to the `Event.write()` method; this includes correcting access to some event attributes elsewhere.
 * This has been covered in https://github.com/QuakeMigrate/QuakeMigrate/commit/b969d0b9b016b2533cab1f1a698ce1291aae4226
- [x] Specify precision for floats in the .event, .amps and .picks files.
 * Covered in https://github.com/QuakeMigrate/QuakeMigrate/commit/739a47881b021957c6cfe11a620cbfa50a5312cc , however the current implementation is inefficient; these precisions should be calculated once for the LUT and stored as an attribute there, rather than once for every event.
- [x] Review the .event file format: which outputs are needed / useful.
 * Covered in https://github.com/QuakeMigrate/QuakeMigrate/pull/76/commits/739a47881b021957c6cfe11a620cbfa50a5312cc
- [x] Review the `QMigrate.io.export` submodule: make modifications for compatibility with new .event file format, and handling of Event attributes; add magnitudes functionality.
* In progress. `to_obspy()` method is covered in https://github.com/QuakeMigrate/QuakeMigrate/pull/76/commits/88768c038a8aaa5fe482bd6fc4cba4452a4f1ea2
* Other export functions will remain a WIP: this will be made clear in the documentation ahead of pushing to master.
- [x] Review soft deprecation for the old .event file format, mainly with respect to the `QMigrate.io.export` submodule
* This will not be included. Users can edit old .event files by simply renaming the columns as specified in https://github.com/QuakeMigrate/QuakeMigrate/pull/76/commits/739a47881b021957c6cfe11a620cbfa50a5312cc

**Improvements still required**
- [x] More efficient implementation of the function to determine the minimum coordinate precision for location outputs in the .event file. Move to LUT. But also ideally maintain backwards compatibility (this may not be possible).
* This will be included in #75 and the current implementation replaced after merging.

**List of major (API-affecting) changes**
* Modified the format of the .event files

PR #79, opened 2020-07-10
Units cleanup

hemmelig:
Bring the `development` branch (specifically those changes made in #76) up to speed with the addition of automatic unit conversion, which was part of the LUT review performed in #75. The determination of the number of decimal places used when writing outputs to file is handled by the `LUT.precision` property, introduced in #75 (specifically 3a94e23).

## Test cases
- [x] test_benchmarks.py reports all example tests pass (strictly speaking, no - but visual comparison between the benchmarks and the example run outputs confirmed that the only changes were to the number of decimal places used in reporting results).

### System details
Ubuntu 18.04

### Final checklist
- [x] All tests still pass.


QMigrate/io/event.py
```diff
@@ -402,20 +402,15 @@ def write(self, run, lut):
 
         # Set floating point precision for locations & loc uncertainties
         ll_corner = lut.coord2grid(lut.ll_corner, inverse=True)
```
twinder36, line 404:
```suggestion
```

Issue #96, opened 2020-09-11
BUG: network fix


tmgreenfield1101:
There was a small bug in the latest changes to allow for networks to be specified. This fixes this bug. Stay tuned for a more complete fix using the full NSLC codes

hemmelig:
I fully agree that adding more information to the input station file is the ideal way to make QM more flexible in terms of the trace meta information. To me, the solution would be to extend (perhaps optionally) the columns that can be provided in the station input file e.g.:

```
Network  |  Code  |  Channels  |  Latitude  |  Longitude  |  Elevation
  bla       bla        bla          bla           bla           bla
```

However, we want to be able to retain the existing behaviour - if only `Code` (or `Name` as it is) is supplied, along with the geographical coordinates, then network and channels are assumed/decided by the information given to archive (in the case of channels)

One major benefit I can see of extending things in this way is that pick files will include all of the important information such that when imported into snuffler they are matched up with the corresponding traces. This doesn't currently really work (and it is a bit annoying!)

tmgreenfield1101:
I can get ahead with stripping out the other updates from the previous PR. I hadn't realised that it held all the baggage from the previous rubbish ones. I'll do better this time. 

My two cents worth is that I think it makes more sense to use trace ids as output from obspy and dealing with internally. This would make QM consistent with accepted naming conventions and include all the necessary information for snuffler. 

I'll make the PR then you can take a look.

hemmelig:
It would be best to start afresh from a new branch off development (so as to include changes made in the intervening time) and submit a new PR.

Again, how does changing things in this way impact previous results? We can construct the obspy-style IDs internally from the information provided in the above format. Doing this in the `read_stations` function and adding a column named `Code` would then minimise the amount of changes needed elsewhere.

Also, we have switched everything over to use `Stream` and `Trace`, rather than ararys of data, so we can use the `.select()` method of Stream to get what we need.

hemmelig:
I think essentially my proposal is the same, but the reverse action is performed in `read_stations`. I think if we do choose to accept it in the `trace_id` style, we would only be able to accept this PR as an addition in a future version OR we would have to ensure backwards-compatibility is maintained. Otherwise, it would have to wait until we have a stable release with a version DOI that does it the old way.

hemmelig:
[...]

PR #100, opened 2020-09-14
Read trace

tmgreenfield1101:
This PR replaces the station specification scheme used by QM. Rather than specifying stations by just their station code. Users use the full NSLC codes used as a default by IRIS. This allows for more flexibility when using waveform archives with more than one network or stations with lots of channel codes. It also updates the current awkward functionality. 

The feature has minimal changes to the code and has been tested using the Reykjanes dataset I am using here. To be clear. the PR here REPLACES the current functionality. If this is accepted as is. All users would have to update their station files to be similar to:
 
Name,Latitude,Longitude,Elevation
RK.ARNA.00.HHZ,63.9062,-22.4173,0.06
RK.ELDV.00.CH[ENZ],63.8656,-22.5026,0.046

Internally the update splits the "Name" into its seperate components. These are then used to request specific channels from the archive.


tmgreenfield1101:
Hi team, 

Could one of you take a look at this as I think this would be very helpful. It builds a lot more flexibility into the system and would solve much of the faff I've had with Dan in selecting the correct data from archives with multiple networks and sampling rates.

Happy if you say it's not good enough, but do need some feedback so I know why.

Tim

hemmelig:
I would be happy to review this change once it has been submitted as a PR isolated from any other changes. Currently this cannot be reviewed and accepted without reviewing and accepting all of the unrelated commits. I've read it through and the idea is reasonable, but I think we can probably come up with a more complete solution to a range of problems by expanding the station file input format.

In general, each pull request should contain a small set of atomic changes to the code that implement a specific feature. In this way we can fully understand the scope of the changes to the base code. The process for this is:

1. Start on the target base branch (at this time `development`, but in future it will either be `master` (for bug fixes etc) or `v1.1` etc (for new features))
2. Create a new branch from this with `git checkout -b <branch-name>` - it would be useful to have a branch name that briefly states the purpose of the branch e.g. `fix_date-time-problems` or `feat_network-information`
3. Run the tests, make sure all benchmarks are reproduced before making any changes
4. Make your changes as a series of logical commits where possible. Your commit message should be as informative as possible. The commit title should follow the `<type>(<scope>): <brief title>` structure, and be no longer than 50 characters. On a newline, write all the information you can.
5. Once all changes are made, re-run all tests. All benchmarks and plots must stay the same, unless as an expected consequence of the changes
6. Once happy, push this branch to remote and open a pull request. Use the PULL_REQUEST_TEMPLATE.md (found on the development branch under `.github` - in future this should come up automatically, but it needs to be on master - soon) and fill in all of the requested information.

Then we can review things in earnest. Much of this information is provided in the CONTRIBUTING.md guide, found on the base branch. I will improve it and make sure it is explicitly linked in the README.md!

hemmelig:
I think best to continue this discussion over at Issue #96 so we can hash out a plan for what changes we would like to make. I would prefer to avoid packing the `Name` column with a lot of information then splitting it apart internally, as it is unclear to me how this would impact work done prior to this change etc.

PR #103, opened 2020-10-14
FEAT - Phase agnosticism

hemmelig:
To free QuakeMigrate from P and S phase gnosticism.

QuakeMigrate will now run on a mixture of P/P and S/S etc and still compute a result. This is useful in scenarios where there is only a single component of data available, or perhaps one component was unavailable during a single timestep - it should remain possible to use the remaining two to build an onset/onsets. Now it is.

Included in this change is a small series of changes that should hopefully make things overall a bit better, including:
- Restructured the Gaussian Picker class to have more atomic behaviour by breaking sections out into independent functions that perform a single, clear action
- Switch to using traces over arrays.
- Reduce redundancy in stacking. We no longer stack useless data! When an onset is unavailable, we simply do not request the corresponding traveltime table and we don't request the onset function. This will at worse leave the computational cost unaffected, but has the potential to have a significant impact on the overall speed and efficiency of running QuakeMigrate, particularly when using evolving networks/networks with very gappy data!
- Slightly changed how the GaussianPicker works - it now uses just the fraction of the phase traveltime to determine the size of the picking window, rather than being based on the marginal window for the event.

## Test cases
Continually ensured that any changes remained backwards compatible (i.e. the benchmark test passed) until absolutely necessary. The only failures are now the availability (which has changed, so will obviously changed) and the picking. S picks are ever so slightly different, likely as a result of the change in how the window is calculated. We no longer us the marginal window at all, but simply tune the picking window size using the fraction of traveltime parameter.

Not finished - new benchmarks to be added when further testing has been performed to ensure this works across all available datasets.

### System details
Ubuntu 18.04, 20.04

### To-do
- [x] Make final benchmarks
- [x] Adjust pick plotting (just write it more explicitly, don't be too clever) - will need to be able to turn off if inappropriate.
- [x] Ensure all documentation is up-to-date.

### Final checklist
- [x] All tests still pass.


quakemigrate/io/data.py
```diff
@@ -578,141 +540,7 @@ def times(self, **kwargs):
         """
 
         # Utilise the .times() method of `obspy.Trace` objects
-        tr = Trace(header={"npts": self.signal.shape[-1],
+        tr = Trace(header={"npts": len(self.waveforms[0].data),
```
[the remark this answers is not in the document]
hemmelig: It's only actually used in plotting and now that most of the data is kept in Trace objects, this is far less necessary. I suspect we can drop it and just fall back on the actual `Trace` routine, rather than 'reconstructing' it.

The full list of uses:
`quakemigrate/plot/event.py` - 2 instances - both should be possible to remove, at (what I expect to be) a negligible cost. Snippet: 
```
    # --- Waveforms ---
    mint, maxt = event.otime - 0.1, event.otime + max_tts*1.5
    for i, station in enumerate(event.data.stations):
        waveforms = event.data.filtered_waveforms.select(station=station)
        for c, comp in zip(WAVEFORM_COLOURS1, "ZNE"):
            data = waveforms.select(channel=f"*{comp}")
            if not bool(data):
                continue
            data = data[0].data
            times_utc = data[0].times(type="UTCDateTime")
            mint_i, maxt_i = [np.argmin(abs(times_utc - t)) for t in (mint, maxt)]
            times_plot = data[0].times(type="matplotlib")[mint_i:maxt_i]
            data[mint_i:]

            # Get station specific range for norm factor
            stat_maxt = event.otime + ttsf[i]*1.5
            norm = max(abs(data[mint_i:np.argmin(abs(times_utc - stat_maxt))]))

            y = data[mint_i:maxt_i] / norm + range_order[i]
            label = f"{comp} component" if i == 0 else None
            ax.plot(times_plot, y, c=c, lw=0.3, label=label, alpha=0.85)
```
There seems to be a random `data[mint_i:]` in there as well that I'm not sure is doing anything...

`quakemigrate/plot/phase_picks.py` - used once. I suspect this can now also be dropped, and simply use:
```
    times = signal[0].times(type="utcdatetime")
```
as this section is done on a station by station basis. Equally, can be moved inside a loop as per the solution above, though it's not perfect as the onset plotting relies on the timestamps being the same as for the data.

`quakemigrate/signal/pickers/gaussian.py` - this is the sticking point. It relies on the timings from the event, there is no other way to get it, currently...

...so, one possible solution is to store onsets as streams, rather than in the dict structure. A little bit of effort to ensure the header is set up, but the serving should be simple enough to adapt, and all the correct timing information is available as the input data to onset is in the form of a `Stream` of `Trace`s. Ultimately this seems at least somewhat sensible, though comes at some extra costs.

quakemigrate/signal/onsets/stalta.py
```diff
@@ -105,30 +106,18 @@ def pre_process(stream, sampling_rate, resample, upfactor, filter_):
...
-        filtered_trace.data = filtered_data
-        filtered_waveforms.append(filtered_trace)
+    # Detrend, apply cosine taper then apply zero-phase band-pass filter
+    # Copy to not operate in-place on the input stream
+    filtered_waveforms = resampled_stream.copy()
+    filtered_waveforms.detrend("linear")
+    filtered_waveforms.detrend("constant")
+    filtered_waveforms.taper(type="cosine", max_percentage=0.05)
+    filtered_waveforms.filter(type="bandpass", freqmin=lowcut, freqmax=highcut,
+                              corners=order, zerophase=True)
```
hemmelig, line 120: We can drop a bunch imports - can't find an easy way to add this as a suggestion.

quakemigrate/signal/pickers/gaussian.py
```diff
@@ -422,19 +431,20 @@ def _find_peak(self, windowed_onset, threshold):
...
         ------
         util.NoOnsetPeak
-            If no onset data exceeds the threshold.
+            If no onset data, or only a single sample, exceeds the threshold.
 
         """
 
         exceedence = np.where(windowed_onset > threshold)[0]
-        if len(exceedence) == 0:
+        if len(exceedence) < 2:
```
hemmelig, line 449: Does this do quite what we want? What if there are multiple periods that exceed the threshold in a window, but the 'true' peak is a single sample? Would the test be better in the `_find_peak` method?

quakemigrate/signal/local_mag/amplitude.py
```diff
@@ -367,7 +373,7 @@ def _get_distances(self, ev_loc, station_data, unit_conversion_factor):
         # Evaulate vertical distance between station and event. Convert to
         # kilometres.
         km_cf = 1000 / unit_conversion_factor
-        z_dist = (evdp - stel) / km_cf
+        z_dist = (evdp - stel) / km_cf # NOTE: stel is actually depth.
```
hemmelig, line 376:
```suggestion
        z_dist = (evdp - stel) / km_cf  # NOTE: stel is actually depth.
```

quakemigrate/signal/local_mag/amplitude.py
```diff
@@ -568,51 +583,48 @@ def _get_amplitude_windows(self, station, i, event, p_ttimes, s_ttimes,
...
-            s_pick = picks.iloc[i*2+1]["ModelledTime"]
+        picks = picks.loc[picks["Station"] == station]
 
-        # Convert to UTCDateTime objects
-        p_pick = UTCDateTime(p_pick)
-        s_pick = UTCDateTime(s_pick)
+        if len(picks) > 0:
+            try:
+                p_pick = picks.loc[picks["Phase"] == "P"]["PickTime"].iloc[0]
+                p_pick = UTCDateTime(str(p_pick))
```
hemmelig, line 616: Out of curiousity, is the cast to `str` necessary?

quakemigrate/signal/local_mag/amplitude.py
```diff
@@ -568,51 +583,48 @@ def _get_amplitude_windows(self, station, i, event, p_ttimes, s_ttimes,
...
+            except (IndexError, ValueError): # UTCDateTime("-1") -> ValueError
```
hemmelig, line 617:
```suggestion
            except (IndexError, ValueError):  # UTCDateTime("-1") -> ValueError
```

quakemigrate/signal/local_mag/amplitude.py
```diff
@@ -568,51 +583,48 @@ def _get_amplitude_windows(self, station, i, event, p_ttimes, s_ttimes,
...
+                p_pick = "-1"
+            try:
+                s_pick = picks.loc[picks["Phase"] == "S"]["PickTime"].iloc[0]
+                s_pick = UTCDateTime(str(s_pick))
+            except (IndexError, ValueError): # UTCDateTime("-1") -> ValueError
```
hemmelig, line 622:
```suggestion
            except (IndexError, ValueError):  # UTCDateTime("-1") -> ValueError
```

hemmelig (commented):
Just one important comment re how we handle single sample spikes in the Gaussian picker and a handful of minor formatting suggestions.

quakemigrate/signal/local_mag/amplitude.py
hemmelig: Possibly something to do with forcing a failure when the pick = -1?

quakemigrate/signal/pickers/gaussian.py
TomWinder: Yep, good catch. Have addressed this in 692a6e7

quakemigrate/signal/local_mag/amplitude.py
TomWinder: Yep. If it reads -1 from the picks dataframe it is an int, which gives a date in 1969 in UTCDateTime. I thought this way gave an easier to read catch (UTCDateTime("-1") --> ValueError --> check isinstance(UTCDateTime))

quakemigrate/plot/event.py
```diff
@@ -148,19 +148,21 @@ def _plot_waveform_gather(ax, lut, event, idx):
     """
 
     # --- Predicted traveltimes ---
-    ttpf, ttsf = [lut.traveltime_to(phase, idx) for phase in "PS"]
-    ttp = [(event.otime + tt).datetime for tt in ttpf]
-    tts = [(event.otime + tt).datetime for tt in ttsf]
-    range_order = abs(np.argsort(np.argsort(ttp)) - len(ttp)) * 2
+    traveltimes = [lut.traveltime_to(phase, idx) for phase in lut.phases]
```
TomWinder, line 151: Will have to re-visit this if we add an extra point to specify phases to be used (e.g. `onset.phases`)

quakemigrate/plot/event.py
```diff
@@ -148,19 +148,21 @@ def _plot_waveform_gather(ax, lut, event, idx):
...
-                   label=f"Modelled {phase}")
+
+    for arrival, c, phase in zip(arrivals, PICK_COLOURS, lut.phases):
+        ax.scatter(arrival, range_order, s=s, c=c, marker="|", zorder=5,
+                   lw=1.5, label=f"Modelled {phase}")
 
     # --- Waveforms ---
     times_utc = event.data.times(type="UTCDateTime")
-    mint, maxt = event.otime - 0.1, event.otime + max_tts*1.5
+    mint, maxt = event.otime - 0.1, event.otime + max(traveltimes[-1])*1.5
```
TomWinder, line 165: Is it necessary to use `traveltimes[-1]`? (I assume this corresponds to the last-added phase / last phase in `lut.phases`, which may not always be the slowest). E.g.:
```    mint, maxt = event.otime - 0.1, event.otime + max([max(tts) for tts in traveltimes])*1.5```

File: quakemigrate/plot/event.py [hunk 10]
```diff
@@ -170,10 +172,9 @@ def _plot_waveform_gather(ax, lut, event, idx):
             if not bool(data):
                 continue
             data = data[0].data
-            data[mint_i:]
 
             # Get station specific range for norm factor
-            stat_maxt = event.otime + ttsf[i]*1.5
+            stat_maxt = event.otime + traveltimes[-1][i]*1.5
```
TomWinder, line 178: Ditto previous comment, though solution here would be different..

TomWinder (commented):
Couple of comments, but looks like it handles everything. Haven't tested yet.

quakemigrate/plot/event.py
hemmelig: It does - yeah, might as well handle in this instance as it is straightforward. I have a feeling there are a few instances where a similar order assumption is used elsewhere in the code, so would need a proper review to make sure.

quakemigrate/plot/event.py
hemmelig: Something like `max(traveltimes[:, i])`

hemmelig: Addressed.

quakemigrate/plot/event.py
hemmelig: Addressed in 1957c29.

quakemigrate/io/data.py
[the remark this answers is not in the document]
TomWinder: So in that first one, `data` is not a `WaveformData` object; just a stream selected from `data.filtered_waveforms` - it's using the `times()` method directly on a trace from that stream, right?

I'm going to leave the `times` method, but just have it operate directly on `data.filtered_waveforms[0]`, without using the `sampling_rate` class variable, which is otherwise unused and unnecessary (and which I am therefore removing).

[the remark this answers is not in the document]
hemmelig: Yes, exactly.

Ok, looks good as a solution for now.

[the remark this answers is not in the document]
TomWinder: Actually, wait. `signal` is also available there (=`event.data.filtered_waveforms`). So can also just use `signal[0].times()` for that case. Unless I'm missing something. In which case I'll just bin the `times` method, because if it's not doing anything extra with the class variables then it's not necessary.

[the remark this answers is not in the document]
TomWinder: Actually actually.. the more I read your summary of the instances of `event.data.times()` the more puzzled I am!

What version of `quakemigrate/plot/event.py` were you looking at when you found those lines? They don't exist in the current one..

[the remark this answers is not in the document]
hemmelig: Oh, my bad I wasn't explicit - that snippet is my proposed solution.

And yes, in that case we should be able to drop the times method entirely.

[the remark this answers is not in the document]
TomWinder: Ha, ok. With you. I think it's all sorted. Pushing shortly (turned into a fairly involved 'clean up')

quakemigrate/io/data.py
```diff
@@ -413,6 +398,92 @@ def __init__(self, starttime, endtime, sampling_rate, stations=None,
...
+            stream (available is again 0 or 1).
+
+        """
+
+        availability = {}
+
+        # Check if any channels in stream
+        if bool(st):
+            # Loop through channels with unique SEED id's
+            for tr_id in sorted([tr.id for tr in st]):
```
hemmelig, line 440: This trace id is just used to drop traces later, right?

quakemigrate/io/data.py
```diff
@@ -413,6 +398,92 @@ def __init__(self, starttime, endtime, sampling_rate, stations=None,
...
+                st_id = st.select(id=tr_id)
+                availability[tr_id] = 1
```
hemmelig, line 442: Perhaps better to set it to 0 at first, drop all the instances of `availability[tr_id] = 0` and set `= 1` as the last line in the loop?

quakemigrate/io/data.py
```diff
@@ -413,6 +398,92 @@ def __init__(self, starttime, endtime, sampling_rate, stations=None,
...
+                        continue
+                # Check data covers full timespan (if requested)
+                if full_timespan:
+                    if len(st_id) > 1:
+                        availability[tr_id] = 0
+                    else:
+                        tr = st_id[0]
+                        if tr.stats.starttime != self.starttime or \
+                            tr.stats.endtime != self.endtime:
+                            availability[tr_id] = 0
```
hemmelig, line 471:
```suggestion
                    if len(st_id) > 1:
                        availability[tr_id] = 0
                    elif st_id[0].stats.starttime != self.starttime or \
                        st_id[0].stats.endtime != self.endtime:
                            availability[tr_id] = 0
```

hemmelig: Also, in what cases does the first test ever fail? Surely if there are multiple streams they must either be overlapping or gapped, hence already handled?

quakemigrate/io/data.py
hemmelig: Similar could be done for `available` - set `= 0` at the top, and only update to 1 if all tests pass.

hemmelig (commented):
Mainly just questions about the availability testing function. I can't see any issues, but there are (possibly) unnecessary redundancies?

quakemigrate/io/data.py
TomWinder: Don't quite follow... I split the stream into traces with unique ID's to enable the 'all_channels' kwarg to work (so if e.g. E is gappy for a station, but N is fine, then available can still = 1), and also in returning the availability as a dict (so that in this situation the information may be used to remove traces from E, if desired -- or fill with zeros, etc.)

quakemigrate/io/data.py
```diff
@@ -413,6 +398,92 @@ def __init__(self, starttime, endtime, sampling_rate, stations=None,
...
+        if bool(st):
+            # Loop through channels with unique SEED id's
+            for tr_id in sorted([tr.id for tr in st]):
+                st_id = st.select(id=tr_id)
+                availability[tr_id] = 1
+
+                # First check data is here:
+                if not bool(st_id):
+                    availability[tr_id] = 0
+                    continue
```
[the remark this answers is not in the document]
TomWinder: Nope, good spot. But raises something I hadn't considered - if only N is fed to this function (in the stream), then it will still report available = 1 even if `all_channels=True`. Will probably fix this by adding an `n_chans` kwarg specifying the number of channels that are expected to be there.

quakemigrate/io/data.py
hemmelig: As in, this trace_id (which presumably is a full SEED ID, right?) is used to identify the traces and to create the 'availability' that is used in `onset` to remove any traces that failed the critera.

hemmelig: * and only used there - but yeah, I understand what it does.

quakemigrate/io/data.py
TomWinder: Gaps only handled already if `allow_gaps=False`

quakemigrate/io/data.py
TomWinder: Yep. It's also used here to get `st_id` - necessary to do by `st.select` because there might be multiple traces with the same id.

quakemigrate/io/data.py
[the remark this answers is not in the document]
TomWinder: (Sorted in 6fe8824)

quakemigrate/plot/phase_picks.py
```diff
@@ -70,87 +74,163 @@ def pick_summary(event, station, signal, picks, onsets, ttimes, window):
...
+        win = windows[ph]
+        onset_max = max(onsets[i][win[0]:win[2]+1])
+        y_max = max(onset_max, thresh)
+        ax.set_ylim(ymin=0, ymax=y_max*1.1)
 
     # --- Plot predicted arrival times ---
-    for i, ax in enumerate(axes):
-        model_pick = otime + ttimes[0] if i % 3 == 0 else otime + ttimes[-1]
+    # Handle case where only a single phase is used
+    ax_ind = [ind for ind in range(5)]
```
hemmelig, line 171: Is the list comprehension necessary? `range` returns an iterator, not a generator, so it doesn't get 'used' up, so I think just `ax_ind = range(5)` should be safe?

hemmelig (commented):
Ok - nothing major, so I think this is all fine. Passes the tests that I tried:

- Normal Iceland icequake run
- Iceland icequakes, just P / just S

quakemigrate/signal/onsets/stalta.py
```diff
@@ -309,7 +314,7 @@ def calculate_onsets(self, data, phases=["P", "S"], log=True):
 
                 # If no data available, skip
                 if available == 0:
-                    logging.debug(f"\t\tNo {phase} onset for {station}.")
+                    logging.info(f"\t\tNo {phase} onset for {station}.")
```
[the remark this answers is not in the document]
hemmelig: e.g. in the Iceland icequake example, you get two messages per timestep for SKG09.

quakemigrate/io/data.py
```diff
@@ -416,6 +416,9 @@ def check_availability(self, st, all_channels=False, allow_gaps=False,
         all_channels : bool, optional
             Whether all supplied channels (distinguished by SEED id) need to
             meet the availability criteria to mark the data as 'available'.
+        n_channels : int, optional
+            If `all_channels=True`, this argument is required (in order to
+            specify the number of channels expected to be present).
```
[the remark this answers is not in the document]
TomWinder: Yes, `all_channels` is off by default.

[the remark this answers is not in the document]
TomWinder: In all but this case, it will replicate previous behaviour if nothing is changed in the run script

quakemigrate/signal/onsets/stalta.py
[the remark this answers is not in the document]
TomWinder: Personally I think it's useful enough to leave in.

quakemigrate/plot/phase_picks.py
TomWinder:
```suggestion
    ax_ind = range(5)
```

setup.py
```diff
@@ -208,13 +208,16 @@ def setup_package():
...
-                     "quakemigrate.export", "quakemigrate.lut",
-                     "quakemigrate.plot", "quakemigrate.signal",
+        "packages": ["quakemigrate",
+                     "quakemigrate.core",
+                     "quakemigrate.export",
+                     "quakemigrate.io",
+                     "quakemigrate.lut",
+                     "quakemigrate.plot",
+                     "quakemigrate.signal",
+                     "quakemigrate.signal.local_mag",
```
hemmelig, line 218:
```suggestion
        "cmdclass": {"build_ext": CustomBuildExt},
        "packages": ["quakemigrate",
                     "quakemigrate.core",
                     "quakemigrate.export",
                     "quakemigrate.io",
                     "quakemigrate.lut",
                     "quakemigrate.plot",
                     "quakemigrate.signal",
                     "quakemigrate.signal.local_mag",
```

hemmelig, line 218:
```suggestion
        "cmdclass": {"build_ext": CustomBuildExt},
        "packages": ["quakemigrate",
                     "quakemigrate.core",
                     "quakemigrate.export",
                     "quakemigrate.io",
                     "quakemigrate.lut",
                     "quakemigrate.plot",
                     "quakemigrate.signal",
                     "quakemigrate.signal.local_mag",
```

quakemigrate/plot/phase_picks.py
```diff
@@ -72,82 +76,165 @@ def pick_summary(event, station, signal, picks, onsets, ttimes, window):
...
 
-    for ax in axes[3:5]:
-        ax.legend()
-    fig.tight_layout(pad=1, w_pad=0)
+    # Add legend
+    for ind in ax_ind:
+        if ind > 2:
+            axes[ind].legend(fontsize=16, loc="upper right")
+
+    fig.tight_layout(pad=4, w_pad=4)
```
hemmelig, line 237:
```suggestion
    fig.tight_layout(pad=1, w_pad=0)
```

hemmelig (commented):
Two points highlighted.

quakemigrate/plot/phase_picks.py
TomWinder, line 237:
```suggestion
    fig.tight_layout(pad=1)
```
1 / 0 is consistent with our other plots, but I think `w_pad=1` is necessary to separate the P and S onset y-axes.

quakemigrate/plot/event.py
```diff
@@ -173,34 +172,42 @@ def _plot_waveform_gather(ax, lut, phases, event, idx):
...
-    mint, maxt = event.otime - 0.1, event.otime + np.max(traveltimes)*1.5
+    waveforms = event.onset_data.filtered_waveforms
+    # Min and max times to plot
+    mint = event.otime - 0.1
+    maxt = min(event.otime + np.max(traveltimes)*1.5, event.data.endtime)
+    # Convert to indices -- will still be the same for sub-sample shifts
+    times_utc = waveforms[0].times("UTCDateTime")
     mint_i, maxt_i = [np.argmin(abs(times_utc - t)) for t in (mint, maxt)]
-    times_plot = event.data.filtered_waveforms[0].times(type="matplotlib")[mint_i:maxt_i]
+    # Loop through stations
```
hemmelig, line 201: Redundant, the next line is pretty explicit.

quakemigrate/plot/event.py
```diff
@@ -173,34 +172,42 @@ def _plot_waveform_gather(ax, lut, phases, event, idx):
...
 
             # Get station specific range for norm factor
             stat_maxt = event.otime + max(traveltimes[:, i])*1.5
-
             norm = max(abs(data[mint_i:np.argmin(abs(times_utc - stat_maxt))]))
 
+            # Generate times for plotting
+            times = tr.times("matplotlib")[mint_i:maxt_i]
+
+            #Plot
```
hemmelig, line 219:
```suggestion
            # Trim to plot limits, normalise, shift by range, then plot
```

quakemigrate/signal/scan.py
```diff
@@ -177,6 +177,9 @@ def __init__(self, archive, lut, onset, run_path, run_name, **kwargs):
             raise util.OnsetTypeError
         self.onset.post_pad = lut.max_traveltime
 
+        self.pre_pad = 0.
+        self.post_pad = 0.
```
hemmelig: Note to review this when the `_read_waveform_data` thing has been sorted out and removed from scan. These ideally shouldn't be exposed/we should avoid passing these around via the object.

quakemigrate/plot/event.py
```diff
@@ -99,7 +99,18 @@ def event_summary(run, event, marginalised_coa_map, lut, xy_files=None):
...
 
-    fig.axes[0].legend(fontsize=14, loc=1, framealpha=1, markerscale=0.5)
+    # Deal with repeating labels in waveform gather legend; combine labels for
+    # ["N", "1"], ["E", "2"]
+    handles, labels = fig.axes[0].get_legend_handles_labels()
+    for cp1, cp2 in [("N", "1"), ("E", "2")]:
+        if all(x in labels for x in [f"{cp1} component",
+                                     f"{cp2} component"]):
+            labels = [f"{cp2}, {cp1} component" if x == f"{cp1} component" \
+                or x == f"{cp2} component" else x for x in labels]
```
hemmelig, line 109:
```suggestion
    for cp1, cp2 in [("N", "1"), ("E", "2")]:
        if all(x in labels for x in [f"{cp1} component", f"{cp2} component"]):
            labels = [f"{cp2}, {cp1} component" if x == f"{cp1} component"
                      or x == f"{cp2} component" else x for x in labels]
```

hemmelig (commented):
Great. No major comments. Being able to (finally) use a subset of stations is a major bonus win!

quakemigrate/plot/event.py
TomWinder:
```suggestion

```

TomWinder (approved):
As far as I'm concerned, this covers everything relevant for this PR. All benchmarks pass for me, and examples run without warnings. What. A. Huge. Amount. Of. Work.

PR #104, opened 2020-10-24
FEAT: Make trigger summary plotting optional

hemmelig:
Make the plotting of trigger summary figures optional. This can be toggled in the same way as the summary figures in QuakeScan - e.g. trigger.plot_trigger_summary = True.

When this option is turned off, no plotting is performed.

When this option is turned on, all plotting is performed and the resultant figure is *always* saved. Previously, it was possible to toggle whether it was saved or not. This would mean running the code twice, even if satisfied after exploring the data in the interactive plot. The keyword argument (provided when running the trig.trigger function) 'savefig' has been changed to 'interactive_plot', as it is far more explicit.

Final change was to add a call to plt.close(fig) at the end of the plotting script, which ensures everything is closed down and there
aren't anymore memory leaks (which can become an issue if running trigger, with plotting on, for an extended period of time).

All example scripts have been updated to reflect the changes.

## Test cases
Tried all combinations of the new parameters. If the old keyword argument `savefig` is used, the script will not run. I think this is fine, personally, but can also handle it. 

### System details
Ubuntu 20.04

### Final checklist
- [x] `development` base branch selected?
- [x] All tests still pass.


TomWinder (changes requested):
I think the premise is good and it's a definite improvement. But I think we should handle savefig - custom exception at the very least, but should be relatively easy to deprecate softly? We're doing this for everything else.

And, more importantly, I think the default for `interactive_plot` should be set to `False` to maintain previous behaviour.

hemmelig:
Easy to handle - add kwargs to the Trigger.trigger method and remap the parameter, if set, to the appropriate new setting and print a message to suggest they update the parameter call.

I do think though that it 'breaks' in an obvious way. I just can't think who this really affects. It adds more work for us than it costs one or two individuals to update their run script.

And yes, I agree best to maintain default behaviour (not plotting interactively) is best. I think I based what I considered 'default' on what the base example scripts do.

TomWinder (approved):
Agreed that it's not worth handling the savefig thing.

PR #107, opened 2020-10-28
fix(locate): catch reading triggered events issue

hemmelig:
Catch and handle (by means of an explicit error message) an issue whereby the user might set an invalid run name or time period for locate, but the script would crash with a fairly impenetrable error.

All previous behaviour identical, now simply checks if any suitable triggered events files were found during the while loop. If not, raise the error. Perhaps it would be worth catching this and printing something neater before exiting the script, but the net effect is the same.

## Test cases
Tested by providing an invalid run name and an invalid time period. An invalid time period is one for which there is no overlap with the time period for which there exists TriggeredEvents.csv files.

### System details
Ubuntu 20.04

### Final checklist
- [x] All tests still pass.

### DIFF
--- a/quakemigrate/util.py
+++ b/quakemigrate/util.py
@@ -549,6 +549,20 @@ class MagsTypeError(Exception):
         super().__init__(msg)
 
 
+class NoTriggerFilesFound(Exception):
+    """
+    Custom exception to handle case when no trigger files are found during
+    locate. This can occur for a number of reasons - an entirely invalid time
+    period was used
+
+    """
+
+    def __init__(self):
+        msg = ("NoTriggerFilesFound: Double check you have supplied a valid "
+               "run name and a time period for which you have run detect.")
+        super().__init__(msg)
+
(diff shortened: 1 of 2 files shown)


quakemigrate/util.py
```diff
@@ -549,6 +549,20 @@ def __init__(self):
         super().__init__(msg)
 
 
+class NoTriggerFilesFound(Exception):
+    """
+    Custom exception to handle case when no trigger files are found during
+    locate. This can occur for a number of reasons - an entirely invalid time
+    period was used
```
hemmelig, line 556:
```suggestion
    locate. This can occur for one of two reasons - an entirely invalid time
    period was used (i.e. one that does not overlap at all with a period of
    time for which there exists TriggeredEvents.csv files) or an invalid run
    name was provided.
```

TomWinder (approved):
All looks good. Tested on Ubuntu 18.04.

Issue #93, opened 2020-07-30
Dealing with different networks


tmgreenfield1101:
QuakeMigrate currently requires all the stations to have unique names. This isn't always the case though, especially when a region has more than one network code. While renaming all the local MiniSEED archive is possible, this shouldn't be a solution as with some large archives this process would be prohibitively slow.

A quick solution is to allow users to use NET.STATION in their input station .csv files and handle this in the archive object. I propose adding a parameter such as `catch_network` to the `Archive` object and adding this catch to a few functions.

For example in the Archive.read_waveform_data function a solution could be:

```

                if self.catch_network and '.' in station:
                    network = station.split('.')[0]
                    station = station.split('.')[1]
                    st_selected += st.select(network=network, 
                                             station=station)
                else:
                    st_selected += st.select(station=station)
```

It makes sense to include this in the PATH_STRUCTURE. i.e.
```

        elif archive_format == "YEAR/JD/NETWORK.STATION*":
            self.format = "{year}/{jday:03d}/{network}.{station}*"
```

The attached file includes these additions
/space/tg286/quake_migrate/QuakeMigrate/quakemigrate/io/data.py

tmgreenfield1101:
finished

PR #110, opened 2020-10-30
feat: added more advanced station naming

tmgreenfield1101:
Added a feature such that stations can be defined in the stations
input file using their full ids (i.e. network, station, location, channel).
This improves the ability for QM users to select station and channel
combinations from larger data archives.

The only significant change to the input file is that users should define
"Station" at minimum in their station input files. QM will then assemble a
naming string using wildards. Users can also add addition columns ("Network",
"Location" and "Channel") which can include regex expressions.

As part of this feature I have removed most of the default naming schemes
(leaving only SeisComp3). Novice users will struggle to find find where
these are defined and I find it difficult to remember the string syntax. I
suggest users instead use the format strings functionality which is more
flexible and easier to understand. But as always, this can be put back in if
necessary

I have tested this on the Reykjanes dataset and it seems to work as expected.

### DIFF
--- a/quakemigrate/io/core.py
+++ b/quakemigrate/io/core.py
@@ -86,14 +86,29 @@ def read_stations(station_file, **kwargs):
 
     stn_data = pd.read_csv(station_file, **kwargs)
 
-    if ("Latitude" or "Longitude" or "Elevation" or "Name") \
+    if ("Latitude" or "Longitude" or "Elevation" or "Station") \
        not in stn_data.columns:
         raise util.StationFileHeaderException
 
     stn_data["Elevation"] = stn_data["Elevation"].apply(lambda x: -1*x)
 
+    if not 'Network' in stn_data.columns:
+        stn_data["Network"] = "*"
+    if not 'Location' in stn_data.columns:
+        stn_data["Location"] = "*"
+    if not 'Channel' in stn_data.columns:
+        stn_data["Channel"] = "*"
+    stn_data['Name'] = stn_data["Network"] + "." + \
+                        stn_data["Station"]
+
     # Ensure station names are strings
+    stn_data = stn_data.astype({"Network": "str"})
+    stn_data = stn_data.astype({"Station": "str"})
+    stn_data = stn_data.astype({"Location": "str"})
+    stn_data = stn_data.astype({"Channel": "str"})
     stn_data = stn_data.astype({"Name": "str"})
+
+    stn_data['Name'] = [s.replace('*', '') for s in stn_data['Name']]
 
     return stn_data
 
--- a/quakemigrate/io/data.py
+++ b/quakemigrate/io/data.py
@@ -39,7 +39,8 @@ class Archive:
         Location of seismic data archive: e.g.: ./DATA_ARCHIVE.
     stations : `pandas.DataFrame` object
         Station information.
-        Columns ["Latitude", "Longitude", "Elevation", "Name"]
+        Columns ["Latitude", "Longitude", "Elevation", "Name", 
+                 "Network", "Station", "Location", "Channel"]
     archive_format : str, optional
         Sets path type for different archive formats.
     kwargs : **dict
@@ -82,10 +83,9 @@ class Archive:
         """Instantiate the Archive object."""
 
         self.archive_path = pathlib.Path(archive_path)
-        self.stations = stations["Name"]
+        self.stations = stations
         if archive_format:
-            channels = kwargs.get("channels", "*")
-            self.path_structure(archive_format, channels)
+            self.path_structure(archive_format)
         else:
             self.format = kwargs.get("format")
 
@@ -107,7 +107,7 @@ class Archive:
 
         return out
 
-    def path_structure(self, archive_format="YEAR/JD/STATION", channels="*"):
+    def path_structure(self, archive_format="YEAR/JD/STATION"):
         """
         Define the path structure of the data archive.
 
@@ -126,20 +126,10 @@ class Archive:
         """
 
         if archive_format == "SeisComp3":
-            self.format = ("{year}/*/{station}/"+channels+"/*.{station}.*.*.D."
+            self.format = ("{year}/{network}/{station}/{channel}.D/{network}.{station}.{location}.{channel}.D."
                            "{year}.{jday:03d}")
-        elif archive_format == "YEAR/JD/*_STATION_*":
-            self.format = "{year}/{jday:03d}/*_{station}_*"
-        elif archive_format == "YEAR/JD/STATION":
-            self.format = "{year}/{jday:03d}/{station}*"
-        elif archive_format == "STATION.YEAR.JULIANDAY":
-            self.format = "*{station}.*.{year}.{jday:03d}"
-        elif archive_format == "/STATION/STATION.YearMonthDay":
-            self.format = "{station}/{station}.{year}{month:02d}{day:02d}"
-        elif archive_format == "YEAR_JD/STATION*":
-            self.format = "{year}_{jday:03d}/{station}*"
-        elif archive_format == "YEAR_JD/STATION_*":
-            self.format = "{year}_{jday:03d}/{station}_*"
+        elif archive_format == "custom":
+            pass
         else:
             raise util.ArchivePathStructureError(archive_format)
 
@@ -233,8 +223,12 @@ class Archive:
             # Re-populate st with only stations in station file, and only
             # data between start and end time needed for QuakeScan
             st_selected = Stream()
-            for station in self.stations:
-                st_selected += st.select(station=station)
+            for n, s, l, c in zip(self.stations['Network'], 
+                                  self.stations['Station'],
+                                  self.stations['Location'],
+                                  self.stations['Channel']):
+                st_selected += st.select(network=n, station=s, 
+                                        location=l, channel=c)
             st = st_selected.copy()
             for tr in st:
                 tr.trim(starttime=starttime, endtime=endtime)
@@ -309,12 +303,16 @@ class Archive:
                                                  dtime=now)
                 files = chain(files, self.archive_path.glob(file_format))
             else:
-                for station in self.stations:
+                for n, s, l, c in zip(self.stations['Network'], 
+                                  self.stations['Station'],
+                                  self.stations['Location'],
+                                  self.stations['Channel']):
                     file_format = self.format.format(year=now.year,
                                                      month=now.month,
                                                      day=now.day,
                                                      jday=now.julday,
-                                                     station=station,
+                                                     network=n, station=s, 
+                                                     location=l, channel=c,
                                                      dtime=now)
                     files = chain(files, self.archive_path.glob(file_format))
             dy += 1
@@ -679,8 +677,11 @@ class WaveformData:
         availability = np.zeros(len(self.stations)).astype(int)
         signal = np.zeros((3, len(self.stations), int(samples)))
 
-        for i, station in enumerate(self.stations):
-            tmp_st = stream.select(station=station)
+        for i, (n, s, l, c) in enumerate(zip(self.stations['Network'],
+                                             self.stations['Station'],
+                                             self.stations['Location'],
+                                             self.stations['Channel'])):
+            tmp_st = stream.select(network=n, station=s, location=l, channel=c)
             if len(tmp_st) == 3:
                 # Check traces are the correct number of samples and not filled
                 # by a constant value (i.e. not flatlines)
--- a/quakemigrate/lut/create_lut.py
+++ b/quakemigrate/lut/create_lut.py
@@ -378,6 +378,7 @@ def _compute_1d_sweep(lut, phase, vmodel, **kwargs):
     (cwd / "model").mkdir(exist_ok=True)
 
     for i, station in enumerate(lut.station_data["Name"].values):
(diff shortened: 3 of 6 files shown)


TomWinder:
As mentioned in the linked issue (#93) a lot of this code is moving in a change that is currently underway. These changes would therefore have to be largely re-written to merge them with what will soon be pushed to development.

On top of that:
* this implementation won't work where the user doesn't specify network, location etc. because `st.select(network="")` etc. returns an empty Stream.
* please do not make changes like removing all the template archive structures without discussing with us first. That may be your opinion - have it how you like on your personal branch - but is not the kind of change that we will make lightly to the public version.
* this doesn't provide any form of deprecation for the previous format of station files (and presumably outputs from pick etc.) - which would be a requirement to incorporate into development.

Overall this PR is a massive improvement on the previous one, but has some fatal flaws. I'm not going to spend any more time on this until the latest changes are merged to development, and we can have a proper think how to add this feature.

tmgreenfield1101:
As I mentioned in the linked issue (#93) I am not aware of where the code is going, but as a start point to including this functionality this a first step and a useful bunch of code for including this. With regards to the raised points


  1.  this does work if the user doesn't specify network, location etc because the read_stations script has been updated to populate these columns if the user does not specify them. In that case the empty column is populated with '*' for exactly the reason you raise.
  2.  Sorry for my ignorance, I had thought these kind of pull requests where the place to discuss such changes. Changes in the PR can be appended to etc without affecting the final code. It's not a big change and can easily be adjusted. In my defence, the fixed formatting strings primarily concern the archive structure we have in place with the Iceland data. Most external users will be using either SDS (the most common presently) or a custom structure which QM is unlikely to support. Which options are present require reading of the source code, which many novice users will be reluctant to search through. I propose it is easier to make the custom strings the default option although I'll admit this doesn't require removing of the current options.
  3.  This is a good point, and I hadn't considered you required backward compatibility here. The update to the station file is only changing "Name" to "Station" and it could be dealt with internally (I think). I'd welcome ideas, but the change is small (consider the changes in GMT versions...). The pickfile is more problematic. Is it ever read by QM?


-------------------------------------------
Tim Greenfield
Leverhulme Early Career Research Fellow
University of Cambridge
tg286@cam.ac.uk
________________________________

quakemigrate/io/core.py
```diff
@@ -86,15 +86,30 @@ def read_stations(station_file, **kwargs):
...
+                        stn_data["Station"]
+
     # Ensure station names are strings
+    stn_data = stn_data.astype({"Network": "str"})
+    stn_data = stn_data.astype({"Station": "str"})
+    stn_data = stn_data.astype({"Location": "str"})
+    stn_data = stn_data.astype({"Channel": "str"})
     stn_data = stn_data.astype({"Name": "str"})
 
+    stn_data['Name'] = [s.replace('*', '') for s in stn_data['Name']]
```
hemmelig, line 111:
```suggestion
    stn_data["TraceID"] = stn_data["TraceID"].apply(lambda x: x.replace("*", ""))
```

quakemigrate/io/data.py
```diff
@@ -39,7 +39,8 @@ class Archive:
         Location of seismic data archive: e.g.: ./DATA_ARCHIVE.
     stations : `pandas.DataFrame` object
         Station information.
-        Columns ["Latitude", "Longitude", "Elevation", "Name"]
+        Columns ["Latitude", "Longitude", "Elevation", "Name", 
+                 "Network", "Station", "Location", "Channel"]
```
hemmelig, line 43: Probably worth indicating in some way that some of these are optional.

quakemigrate/io/data.py
```diff
@@ -126,20 +126,10 @@ def path_structure(self, archive_format="YEAR/JD/STATION", channels="*"):
...
-        elif archive_format == "YEAR/JD/STATION":
-            self.format = "{year}/{jday:03d}/{station}*"
-        elif archive_format == "STATION.YEAR.JULIANDAY":
-            self.format = "*{station}.*.{year}.{jday:03d}"
-        elif archive_format == "/STATION/STATION.YearMonthDay":
-            self.format = "{station}/{station}.{year}{month:02d}{day:02d}"
-        elif archive_format == "YEAR_JD/STATION*":
-            self.format = "{year}_{jday:03d}/{station}*"
-        elif archive_format == "YEAR_JD/STATION_*":
-            self.format = "{year}_{jday:03d}/{station}_*"
```
hemmelig, line 142: No need to hard deprecate these. Maybe add some logged statements here saying how to update the run scripts ahead of deprecation in the future.

quakemigrate/io/data.py
```diff
@@ -126,20 +126,10 @@ def path_structure(self, archive_format="YEAR/JD/STATION", channels="*"):
...
+        elif archive_format == "custom":
+            pass
```
hemmelig, line 132: Why include this at all?

quakemigrate/io/data.py
```diff
@@ -107,7 +107,7 @@ def __str__(self):
 
         return out
 
-    def path_structure(self, archive_format="YEAR/JD/STATION", channels="*"):
+    def path_structure(self, archive_format="YEAR/JD/STATION"):
```
hemmelig, line 110: The default here needs to change if we remove the option it corresponds to.

quakemigrate/io/data.py
```diff
@@ -126,20 +126,10 @@ def path_structure(self, archive_format="YEAR/JD/STATION", channels="*"):
         """
 
         if archive_format == "SeisComp3":
-            self.format = ("{year}/*/{station}/"+channels+"/*.{station}.*.*.D."
+            self.format = ("{year}/{network}/{station}/{channel}.D/{network}.{station}.{location}.{channel}.D."
```
hemmelig, line 129:
```suggestion
            self.format = ("{year}/{network}/{station}/{channel}.D/{network}."
                           "{station}.{location:02d}.{channel}.D.{year}."
                           "{jday:03d}")
```
Failsafe.

quakemigrate/io/data.py
```diff
@@ -126,20 +126,10 @@ def path_structure(self, archive_format="YEAR/JD/STATION", channels="*"):
...
                            "{year}.{jday:03d}")
```
hemmelig, line 130:
```suggestion
```

quakemigrate/io/data.py
```diff
@@ -233,8 +223,12 @@ def read_waveform_data(self, starttime, endtime, sampling_rate, pre_pad=0.,
...
             # data between start and end time needed for QuakeScan
             st_selected = Stream()
-            for station in self.stations:
-                st_selected += st.select(station=station)
+            for n, s, l, c in zip(self.stations['Network'], 
+                                  self.stations['Station'],
+                                  self.stations['Location'],
+                                  self.stations['Channel']):
+                st_selected += st.select(network=n, station=s, 
+                                        location=l, channel=c)
```
hemmelig, line 231:
```suggestion
            for _, station in self.stations.iterrows():
                st_selected += st.select(network=station["Network"],
                                         station=station["Station"],
                                         location=station["Location"],
                                         channel=station["Channel"])
```

quakemigrate/io/data.py
```diff
@@ -309,12 +303,16 @@ def _load_from_path(self, starttime, endtime):
...
+                                  self.stations['Station'],
+                                  self.stations['Location'],
+                                  self.stations['Channel']):
                     file_format = self.format.format(year=now.year,
                                                      month=now.month,
                                                      day=now.day,
                                                      jday=now.julday,
-                                                     station=station,
+                                                     network=n, station=s, 
+                                                     location=l, channel=c,
```
hemmelig, line 315: Same as previous suggestion - better to just loop over the rows and extract each thing explicitly. There's absolutely no need to use single letters for variables and should be avoided as much as possible.

quakemigrate/io/data.py
```diff
@@ -679,8 +677,11 @@ def _station_availability(self, stream):
         availability = np.zeros(len(self.stations)).astype(int)
         signal = np.zeros((3, len(self.stations), int(samples)))
 
-        for i, station in enumerate(self.stations):
-            tmp_st = stream.select(station=station)
+        for i, (n, s, l, c) in enumerate(zip(self.stations['Network'],
+                                             self.stations['Station'],
+                                             self.stations['Location'],
+                                             self.stations['Channel'])):
+            tmp_st = stream.select(network=n, station=s, location=l, channel=c)
```
hemmelig, line 684: Same again.

quakemigrate/lut/create_lut.py
```diff
@@ -378,6 +378,7 @@ def _compute_1d_sweep(lut, phase, vmodel, **kwargs):
     (cwd / "model").mkdir(exist_ok=True)
 
     for i, station in enumerate(lut.station_data["Name"].values):
+        # station = station.replace('*', '')
         logging.info(f"\t\t...running NonLinLoc - station: {station:5s} - "
                      f"{i+1} of {stations_xyz.shape[0]}")
```
hemmelig, line 383:
```suggestion
    for i, station in lut.station_data.iterrows():
        logging.info(f"\t\t...running NonLinLoc - station: {station.TraceID:5s} - "
                     f"{i+1} of {stations_xyz.shape[0]}")
```

quakemigrate/lut/create_lut.py
```diff
@@ -405,7 +406,7 @@ def _compute_1d_sweep(lut, phase, vmodel, **kwargs):
         to_read = cwd / "time" / f"layer.{phase}.{station}.time"
         gridspec, _, traveltimes = _read_nlloc(to_read, ignore_proj=True)
 
-        lut.traveltimes.setdefault(station, {}).update(
+        lut.traveltimes.setdefault(lut.station_data["Name"][i], {}).update(
```
hemmelig, line 409:
```suggestion
        lut.traveltimes.setdefault(station.TraceID, {}).update(
```

quakemigrate/plot/event.py
```diff
@@ -191,7 +191,7 @@ def _plot_waveform_gather(ax, lut, event, idx):
     ax.set_ylim([0, max(range_order)+2])
     ax.xaxis.set_major_formatter(util.DateFormatter("%H:%M:%S.{ms}", 2))
     ax.yaxis.set_ticks(range_order)
-    ax.yaxis.set_ticklabels(event.data.stations, fontsize=14)
+    ax.yaxis.set_ticklabels(event.data.stations['Name'], fontsize=14)
```
hemmelig, line 194:
```suggestion
    ax.yaxis.set_ticklabels(event.data.stations["TraceID"], fontsize=14) 
```

quakemigrate/signal/local_mag/amplitude.py
```diff
@@ -287,7 +293,7 @@ def get_amplitudes(self, event, lut):
                     and tr[0].stats.endtime > (tr_end - tr[0].stats.delta):
                     tr = tr[0]
                 else:
-                    amps[0] = f".{station}..{comp}"
+                    amps[0] = f"{network}.{station}.{location}.{comp}"
```
hemmelig, line 296:
```suggestion
                    amps[0] = f"{network}.{station}.{location:02d}.{comp}"
```
Failsafe.

quakemigrate/signal/local_mag/amplitude.py
```diff
@@ -262,7 +262,10 @@ def get_amplitudes(self, event, lut):
 
         # Loop through stations, calculating amplitude info
         for i, station_data in lut.station_data.iterrows():
-            station = station_data["Name"]
+            network = station_data["Network"]
+            station = station_data["Station"]
+            location = station_data["Location"]
+            channel = station_data["Channel"]
```
hemmelig, line 268: Is there any need to unpack here, or can we just have `station_data.Network` when `network` is needed, etc.

quakemigrate/signal/pickers/gaussian.py
```diff
@@ -464,7 +464,8 @@ def plot(self, event, lut, picks, ttimes, run):
                 ax.legend(fontsize=14)
 
             fstem = f"{event.uid}_{station}"
-            file = (fpath / fstem).with_suffix(".pdf")
+            # file = (fpath / fstem).with_suffix(".pdf")
+            file = fpath / (fstem + '.pdf')
```
hemmelig, line 468:
```suggestion
            file = fpath / (fstem + ".pdf")
```

hemmelig (changes requested):
I echo Tom - this is good in principle, but will need to be carefully reviewed in light of the changes to how waveform data is handled. I've added some comments and suggestions throughout to give an _idea_ of what might need to be changed currently.

Don't commit the suggestions individually - these can be batched together by going to the file tab and starting a batch commit. But this should wait until the aforementioned changes have been implemented.

Straight away I can see that the icequake examples probably won't work - they are there for a reason, as is the benchmark test. For example, the changes only seem to have been made for when the NonLinLoc LUT is being created.

hemmelig:
Just a comment on backwards-compatibility - in the interest of being as reproducible as possible, we want to 1. ensure changes are carefully tracked and versions are updated appropriately and 2. softly deprecate anything important i.e. trying wherever possible to retain backwards-compatibility, warning users of upcoming changes, and trying to ensure something someone ran 1 week/month/year ago still works exactly (or as close to exactly) the same, if possible. Step changes should only be to handle true bugs.

PR #111, opened 2020-10-30
bugfix(plot/trigger.py): fix invisible coalescence

tmgreenfield1101:
Fixed the plot/trigger.py script so it is visible when
you select savefig=False.

### DIFF
--- a/quakemigrate/plot/trigger.py
+++ b/quakemigrate/plot/trigger.py
@@ -82,7 +82,7 @@ def trigger_summary(events, starttime, endtime, run, marginal_window,
 
     dt = pd.to_datetime(data["DT"].astype(str)).values
 
-    fig = plt.figure(figsize=(30, 15))
+    fig = plt.figure(figsize=(15, 7.5))
     gs = (9, 18)
 
     # Create plot axes, ordering: [COA, COA_N, AVAIL, XY, XZ, YZ]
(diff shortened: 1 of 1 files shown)


TomWinder:
Thanks for raising this: but again, it is most useful to submit it as an issue for discussion first.

We have considered this previously, and unfortunately it is not as simple an issue to deal with as it first appears. The rendering of the interactive plots is monitor / operating system / software dependent, so to find a set of plotting parameters which work generally is far more difficult than for the plots saved as PDF's.

The change you have proposed presumably makes the interactive plots look nice on your monitor / system etc., but at the cost of making changes to the saved PDF's, which we do not wish to make at this time, and which should be the primary output used to view the trigger results. The longer-term solution will be to package the plotting options in user-facing rcparams, which can then be edited (perhaps even on a per-plot-type basis) by each user individually to suit their use-case. However, this is an involved process. We are making the first step on this journey in a current PR #104 by (at least superficially) separating the interactive plot from the saved PDF. However the full solution will take some time.

In the meantime we think using the saved PDF's is an acceptable workaround. It is also perhaps best to switch the examples to not plot the interactive version for the time being - thoughts @hemmelig?

Couple more general pieces of feedback: please use the pull request template (https://github.com/QuakeMigrate/QuakeMigrate/blob/development/.github/PULL_REQUEST_TEMPLATE.md) to help with providing a useful description of the PR, and please check through the diff before submitting to help avoid cluttering the code with obsolete commented out lines.

hemmelig:
I agree with switching default behaviour of examples to plot the pdfs, not interactive.

tmgreenfield1101:
If switching to pdf as the default is preferred (which I think is right). I would suggest either removing the functionality of the interactive pdf or updating it so it actually works on common machines without users having to make special "rcparams" like matplotlib environments. 

Presently, users on every computer I've used cannot use the interactive plot to zoom in on the figure as the line width is too small. Novice users especially should be encouraged to look at the trigger summary plot which contains much The simple additions I've included here change the final pdf only in such the final pdf is physically smaller.

PR #113, opened 2020-12-22
[FIX]: Amplitudes fixes

TomWinder:
To fix a couple of issues in the local_mag.amplitudes sub-module:
- [x] The calculation of the pre-pad and post-pad needed to ensure the taper applied during response removal etc. didn't encroach on the noise / signal amplitude measurements was broken
- [x] To improve the overlap handling where the P and S signal windows overlap.

## Test cases
* Padding-related changes tested by running the dike intrusion example. All amplitude measurements are identical in most cases, except where the taper was previously affecting the result. E.g. for one channel on station FLUR for event 20140824000109120 where the noise window was previously slightly affected by the taper (the new noise measurement is 0.087388; previously was 0.087386).
* Tested the dike intrusion example with a pre_cut and post_cut specified.
* Tested the Iceland Icequake example with a pre_cut and/or post_cut specified.
* NOTE! a bug in write_cut_waveforms was uncovered during this testing -- the user-specified pre_cut and post_cut were not being passed on; plus a minor bug in cut_waveforms.py. They have also been fixed.
* Window related changes also tested by running the dike intrusion example. There are small differences to most signal amplitude measurements (and to a lesser extent their peak times, and approx peak frequencies). A thorough inspection of the differences suggest everything is working sensibly and as expected.

--> NO! small changes to the start and end times of cut waveforms for the dike example (no manual pre_cut and post_cut are set), and small changes to amplitude measurements as discussed above. Once these changes have been checked and approved in principal I will update the benchmarks.
--> All pass now updated.

### System details
Ubuntu 18.04

### Final checklist
- [x] benchmarks updated
- [x] All tests still pass. -- NO! need updating when changes are approved.


quakemigrate/signal/local_mag/amplitude.py
```diff
@@ -585,13 +590,11 @@ def _get_amplitude_windows(self, station, i, event, p_ttimes, s_ttimes,
...
-                windows = [[p_start, p_end],
-                           [p_end, s_end]]
-            else:
-                windows = [[p_start, s_pick - event.marginal_window / 2],
-                           [s_pick - event.marginal_window / 2, s_end]]
+            mid_time = p_end + (s_start - p_end) / 2
+            windows = [[p_start, mid_time],
+                       [mid_time, s_end]]
         else:
             windows = [[p_start, s_start],
```
hemmelig, line 596: Is this intentional? Rather than `p_end`?

quakemigrate/signal/local_mag/amplitude.py
```diff
@@ -600,8 +603,9 @@ def _get_amplitude_windows(self, station, i, event, p_ttimes, s_ttimes,
 
     def _get_picks(self, station, event):
         """
-        Get picks from this station for this event. If no phase picks are
-        found, or if no pick was found, -1 is returned.
+        Get picks from this station for this event. If no phase picks is
```
hemmelig, line 606:
```suggestion
        Get picks from this station for this event. If no phase pick is
```

hemmelig (approved):

quakemigrate/signal/local_mag/amplitude.py
TomWinder: Yep, no real reason not to extend the P window as far as possible. The windows thing is all quite ambiguous; there doesn't seem to be any consensus on a standard practice, particularly for P.

PR #114, opened 2020-12-23
[FIX]: Fix resampling/upsampling bug; more explicit handling of data timing.

TomWinder:
Previously when data was upsampled during resampling (e.g. if going from raw data sampling rate of 40 Hz to an onset function sampling rate of 50 Hz), no check was made that the start and endtimes of the data read from the archive exactly matched the start and endtimes requested (which are calculated assuming the data being read is at the onset function sampling rate, or a multiple if it). In many cases this would not be the case, leading to this data being needlessly discarded.

In this PR this bug is fixed, along with:
* Adding a couple more options to `data.check_availability()` to explicitly check that the data is at the correct sampling rate, and an option to more strictly check it covers the requested timespan (by checking the start and endtimes, not just the number of samples).
* Added a function to explicitly handle "off-sample" timestamped data. See commit https://github.com/QuakeMigrate/QuakeMigrate/commit/b2aafe926d69b231eb172ccc1e8bedfb24e83cd5 for a full explanation. Previously this was implicitly shifted - now there is an explicit option to either continue this previous behaviour, or interpolate the data to shift the sample timestamps without affecting the timing.
* Making clearer when we are using `nearest_sample=True` for obspy.read(), obspy.stream.trim() and why

## Test cases
* The resampling/upsampling fix has been tested with data which was previously affected by this bug (40 Hz data being resampled to 50 Hz onset function sampling rate) and shows all tests now pass as expected. Plus a thorough inspection of the new debug messages added to the code.
* The shift_to_sample function has been tested by artificially shifting the timestamps of 50 Hz data by 0.003, 0.01 and 0.011 seconds using `interpolate=False`. All tests produce the expected results.
* The previous test was also run with a time shift of 0.011 s and `interpolate=True` and produced the expected results compared to the non-shifted data and to the shifted data run with `interpolate=False`. This was primarily compared by looking at the difference in the pick times.

### System details
Ubuntu 18.04

### Final checklist
- [x] All tests still pass.

### DIFF
--- a/quakemigrate/io/data.py
+++ b/quakemigrate/io/data.py
@@ -396,7 +396,8 @@ class WaveformData:
 
     def check_availability(self, st, all_channels=False, n_channels=None,
                            allow_gaps=False, full_timespan=True,
-                           check_sampling_rate=False, sampling_rate=None):
+                           check_sampling_rate=False, sampling_rate=None,
+                           check_start_end_times=False):
         """
         Check waveform availability against data quality criteria.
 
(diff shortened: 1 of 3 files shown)


hemmelig (commented):
Approved, pending a quick test against the off-sample MetMalaysia data.

hemmelig (approved):
After fixes to address negative time-shifts in data, all seems to be in order. Approved.

PR #115, opened 2020-12-23
[FEAT]: Set value of STA/LTA onset in taper regions to 0

TomWinder:
The data read from the archive is padded to allow for the pre- and post-pad needed for the particular onset function algorithm, plus an additional padding to account for the region of data that will be tapered when applying filters etc. Though implicitly disregarded during migration, this region of the onset functions is still used in calculating a pick threshold in the GaussianPicker. The erroneously high onset function values at the start of the onset function (where the taper leads to a high values of the STA/LTA) can therefore interfere with setting an appropriate pick threshold. Here a function is added to set the value of the onset functions to 0 within this time window.

## Test cases
Tested on all of the examples, plotting the entire time window of the onset function & waveforms on the pick plots to ensure the behaviour is as expected. The examples all run smoothly, but the benchmarks will now NOT pass, due to the pick thresholds, and to a minor extent the pick times, uncertainties and SNR's of existing picks, having changed. Once these changes have been approved, the benchmarks will be updated accordingly. The picking performance, as expected, is considerably improved.

- [x] test_benchmarks.py reports all example tests pass -- [NO!] -- Yes, now benchmarks have been updated.

### System details
Ubuntu 18.04

### Final checklist
- [x] All tests still pass. -- [NO!] -- yes, now benchmarks have been updated.


hemmelig (commented):
Looks good. Again, approved pending testing against the Kinabalu example.

hemmelig:
All works for the Kinabalu dataset. Approved.

hemmelig (approved):

PR #121, opened 2021-01-12
FIX: Minor bug fixes

TomWinder:
To fix two bugs, and alter the handling of PickOrderException's during amplitude measurement:
1. Handle amplitude measurements outside the frequency range where the filter gain is defined (possible due to the very approximate estimate of the frequency of the peak-to-peak amplitude measurement)
2. Correct the handling of ResponseNotFoundError's during output of real or Wood-Anderson simulated cut waveforms / add handling for the AttributeError raised if a response inventory hasn't been provided.
3. Emit a warning if a PickOrderException occurs during amplitude measurement, instead of raising.

### Relevant Issues
N/A

## Test cases
(1) Tested on data which had initially exposed these issues (Askja 2007-2020 dataset), and (2) tested on the dike example with mags switched on/off and response inventory (not) provided.

### System details
Ubuntu 18.04 // python 3.8

### Final checklist
- [x] `master` base branch selected?
- [x] All tests still pass.

### DIFF
--- a/quakemigrate/signal/local_mag/amplitude.py
+++ b/quakemigrate/signal/local_mag/amplitude.py
@@ -325,11 +325,14 @@ class Amplitude:
                 else:
                     filter_sos = None
 
-                windows, picked = self._get_amplitude_windows(station, i,
-                                                              event, p_ttimes,
-                                                              s_ttimes,
-                                                              lut.fraction_tt)
-                amps[14] = picked
+                try:
+                    windows, picked = self._get_amplitude_windows(station, i,
+                        event, p_ttimes, s_ttimes, lut.fraction_tt)
+                    amps[14] = picked
+                except util.PickOrderException as e:
+                    logging.warning(f"{e}")
+                    amplitudes.loc[i*3+j] = amps
+                    continue
 
                 amps = self._measure_signal_amps(amps, tr, windows,
                                                  self.noise_measure,
--- a/quakemigrate/signal/local_mag/amplitude.py
+++ b/quakemigrate/signal/local_mag/amplitude.py
@@ -770,8 +770,14 @@ class Amplitude:
                 _, filter_gain = sosfreqz(filter_sos, worN=[approx_freq],
                                           fs=tr.stats.sampling_rate)
                 filter_gain = np.abs(filter_gain[0])
-                half_amp /= filter_gain
-                average_amp /= filter_gain
+                if not filter_gain:
+                    logging.info("\t    Warning: Invalid frequency ("
+                                 f"{approx_freq:.5g} Hz) for {phase}_amp "
+                                 f"measurement on:\n\t\t{tr}")
+                    continue
+                else:
+                    half_amp /= filter_gain
+                    average_amp /= filter_gain
 
             # Put in relevant columns for P / S amplitude, approx_freq,
             # p2t_time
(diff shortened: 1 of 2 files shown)


quakemigrate/signal/local_mag/amplitude.py
```diff
@@ -770,8 +770,14 @@ def _measure_signal_amps(self, amps, tr, windows, method="RMS",
                 _, filter_gain = sosfreqz(filter_sos, worN=[approx_freq],
                                           fs=tr.stats.sampling_rate)
                 filter_gain = np.abs(filter_gain[0])
-                half_amp /= filter_gain
-                average_amp /= filter_gain
+                if not filter_gain:
```
hemmelig, line 776: What does `np.abs()` of a (presumably) `None` return? Still `None`?

quakemigrate/signal/local_mag/amplitude.py
```diff
@@ -325,11 +325,14 @@ def get_amplitudes(self, event, lut):
...
                     filter_sos = None
 
-                windows, picked = self._get_amplitude_windows(station, i,
-                                                              event, p_ttimes,
-                                                              s_ttimes,
-                                                              lut.fraction_tt)
-                amps[14] = picked
+                try:
+                    windows, picked = self._get_amplitude_windows(station, i,
+                        event, p_ttimes, s_ttimes, lut.fraction_tt)
```
hemmelig, line 330:
```suggestion
                    windows, picked = self._get_amplitude_windows(
                        station, i,
                        event, p_ttimes, s_ttimes, lut.fraction_tt)
```

quakemigrate/signal/local_mag/amplitude.py
TomWinder: See commit message - it returns `0.`, not `None`

hemmelig (approved):
All looks good - ran tests and no issues uncovered. All changes look good to me.

quakemigrate/signal/local_mag/amplitude.py
TomWinder: Not this?
```suggestion
                    windows, picked = self._get_amplitude_windows(
                        station, i, event, p_ttimes, s_ttimes, lut.fraction_tt)
```

TomWinder: Checked on an online tool and this passes the checks (need to install flake8 locally..)

hemmelig: Yep, that's also fine! My b.

quakemigrate/signal/local_mag/amplitude.py
hemmelig: So the test is using the falsity of `0` - can anything else be that would trigger this (e.g. a `None`) for which we wouldn't want the same message? I can't think of one but just in case.

TomWinder: I chose this test to also cover `None` or `NaN` - not immediately obvious how they would come about, but they would also lead the following lines to break. Potentially in that case a different message would be appropriate, but these are hypotheticals of hypotheticals - the printout of the invalid frequency should show if something different is going wrong.

hemmelig: Aye, ok, all good.

hemmelig (approved):
Re-approving.

PR #122, opened 2021-01-12
Small set of changes ahead of official v1.0 release

hemmelig:
Introduces Continuous Integration (Travis CI) for the project, which tests the package against Python 3.6, 3.7, and 3.8 as well as against Linux (Xenial) and MacOSX.

Also includes:

- Update to copyright notices (99868e3)
- Pin pandas version to ensure old versions are updated (c78e45e)
- Prep. for pip installation (b2e7391)
- Add badges! Exciting! (194297f

### Final checklist
- [x] All tests still pass.


tests/test_benchmarks.py
```diff
@@ -31,30 +38,51 @@ def test_lut(self):
...
+            print("\t\t...'grid_proj'...")
+            self.assertEqual(b_lut.grid_proj, t_lut.grid_proj)
+            print("\t\t...'coord_proj'...")
+            self.assertEqual(b_lut.coord_proj, t_lut.coord_proj)
+            print("\t\t...'ll_corner'...")
+            self.assertTrue(np.isclose(b_lut.ll_corner, t_lut.ll_corner).all())
+            print("\t\t...'ur_corner'...")
+            self.assertTrue(np.isclose(b_lut.ur_corner, t_lut.ur_corner).all())
+            print("\t\t...'node_spacing'...")
+            self.assertTrue(np.equal(b_lut.node_spacing, t_lut.node_spacing).all())
```
hemmelig, line 74:
```suggestion
             self.assertTrue(np.equal(b_lut.node_spacing,
                                      t_lut.node_spacing).all())
```

hemmelig:
Last thing to add is the Changelog.

.travis.yml
```diff
@@ -0,0 +1,129 @@
...
+      os: osx
+      language: generic
+      env: PYTHON_VERSION=3.8
+
+sudo: false
+
+before_install:
+  # Make Python scripts executable
+  - chmod +x tests/*
+  - chmod +x examples/Icequake_Iceland/*.py
```
TomWinder, line 40: Necessary? Presumably not if the dike example works without this

.travis.yml
```diff
@@ -0,0 +1,129 @@
...
+  - conda config --set always_yes yes --set changeps1 no
+  - conda config --add channels conda-forge
+  - conda update -q conda
+  - conda info -a
+  - echo $PYTHON_VERSION
+  - conda create -q -n test-environment python=$PYTHON_VERSION coverage flake8 matplotlib mock numpy obspy pandas pyproj scipy
+  - source activate test-environment
+
+  # Enforce single-threaded just to stay in memory
+  - export OMP_NUM_THREADS=1;
```
TomWinder, line 65: I think this does nothing: these variables are already set in each example script that is run, and will be over-ridden by the `scan.threads` specified in the run script.

tests/test_import.py
```diff
@@ -0,0 +1,61 @@
...
+
+class TestImport(unittest.TestCase):
+    def test_import(self):
+        i = 0
+        import sys
+        if sys.version_info.major != 3:
+            print("QuakeMigrate does not support Python 2.x")
+            i += 1
+        if sys.version_info.minor < 6:
+            print("QuakeMigrate only supports Python 3.5 and up.")
```
TomWinder, line 25:
```suggestion
            print("QuakeMigrate only supports Python 3.6 and up.")
```

TomWinder (changes requested):
Overall looks good. Strictly it should be "2020-2021" (no spaces) which bugs me but obviously doesn't actually matter. Exciting to have CI up and running.

Request changes is purely for python version printout in test_import.py

.travis.yml
TomWinder: Nope, they're not executed. It's just python that needs permissions - so I'm surprised if `python tests/test_benchmarks.py` doesn't work without this being set.

TomWinder:
> Last thing to add is the Changelog.

Will include this in a following PR.

TomWinder (approved):
Approved

PR #127, opened 2021-02-02
FEAT: Preparations for PyPI and pip install

hemmelig:
## What is the purpose of this Pull Request? 
Prepare everything required for the registration of QuakeMigrate with
the Python Package Index (PyPI). This will enable users to install the
package using pip as:

`pip install quakemigrate`

The following changes were required:

- Added a MANIFEST.in file that specifies the header files for the C
extension
- Corrected the README.md and docs/index.rst files
- Added Python version requirements. Must be using 3.6, 3.7, or 3.8.
Things do not currently work with 3.9 as there are no wheels available
for some of the specific versions, namely matplotlib 3.2.2. ObsPy is
also yet to be fully available for Python 3.9
- Adjusted setup.py to properly handle the requirement for numpy to be
available during the installation process.
- Added explicit requirement for PyQt5. This is used as the matplotlib
backend, which has been the case for much of the development. We felt it
safer to just add this as a requirement for now, and investigate whether
or not it is strictly required later on.

Work remaining:
- Need to figure out how to provide binary distributions ('wheels')
which contain a precompiled version of the package that should help us
bypass some of the difficulties we face regarding Operating Systems,
compilers, etc etc...
- Update documentation to point users to this easier installation process!

The nice thing about the installation via pip is that we are only then making the user download the bare minimum (merely some kB!), rather than all the 'overhead' that comes with the installation from GitHub source. 

## Test cases
Ran extensive testing using the test version of the Python Package Index (test.pypi.org), which is an entirely independent index that can be used to iron out the issues with registering a package on the main Python Package Index. Building of the source distribution was done in my main conda environment for running QM.

- [x] All examples run without any new warnings
- [x] test_benchmarks.py reports all example tests pass

### System details
Ubuntu 20.04

### Final checklist
- [x] All tests still pass.
- [ ] Update documentation to cover the installation through pip process

PR #138, opened 2022-06-02
FEAT: wheel building and PyPI releases

hemmelig:
Prepare everything required for the registration of QuakeMigrate with
the Python Package Index (PyPI). This will enable users to install the
package using pip as:

`pip install quakemigrate`

This is achieved using a GitHub Action that simulates the various supported platforms, compiles and links the C code, and builds the package into a simple downloadable "wheel". This skips the requirement for users to install a C compiler themselves and also minimises the download overhead! This PR supersedes PR #127.

A new commit being pushed to either main or a branch with a name conforming to `test_wheels_*` will trigger a new build. New tags will also be uploaded to TestPyPI. Full releases of the main branch will result in wheels being built and then uploaded to PyPI - they can then be installed via `pip`!

The following changes were required:

- add `.github/workflows/build_wheels.yml` file that guides the GitHub Action
- Major overhaul of the setup.py script to be better organised and much more readable
- Corrected the `README.md` and docs/index.rst files
- Added Python version requirements. Must be using 3.7, 3.8, or 3.9.
Things do not currently work with 3.10 as there are no wheels available
for some of the specific versions.
- Adjusted `setup.py` to properly handle the requirement for numpy to be
available during the installation process.
- Added explicit requirement for PyQt5. This is used as the matplotlib
backend, which has been the case for much of the development. We felt it
safer to just add this as a requirement for now, and investigate whether
or not it is strictly required later on.

The nice thing about the installation via pip is that we are only then making the user download the bare minimum (merely some kB!), rather than all the 'overhead' that comes with the installation from GitHub source. 

## Test cases
Ran extensive testing using the test version of the Python Package Index (test.pypi.org), which is an entirely independent index that can be used to iron out the issues with registering a package on the main Python Package Index. Building of the source distribution was done in my main conda environment for running QM.

### System details
Ubuntu 20.04

### Final checklist
- [x] All tests still pass.
- [x] Update documentation to cover the installation through pip process


hemmelig:
There is some major faff with how best to install the Qt backend (`pip install PyQt5` vs `conda install pyqt`) which results in some potentially 'catastrophic' problems with anaconda.

See here for more information:
https://github.com/ContinuumIO/anaconda-issues/issues/1554

The below outlines what is likely to be the best approach:
https://github.com/biolab/orange3/pull/5566

Briefly - _remove_ the explicit PyQt5 dependency from `setup.py`, `requirements.txt` and instead, if the user does not have a suitable version of Qt somewhere, catch the import error and throw a message prompting the user to install it.

TomWinder:
Nice - I think this is a good stop-gap solution. Couple of things:
* We will need to do something about the required Travis check - possibly can be overridden.
* I think best to keep the instructions for creation of an environment within the "from source" instructions -- makes much more sense here, given they are cloning the repo and therefore have the environment.yml locally. So (1) clone (2) use .yml to make environment; activate then (3) pip install . (or whatever)
* I wonder whether it's worth retaining a couple of the optional extras for the pip installation instructions - i.e. in the environment file we have ipython and jupyter for convenience; not sure it does any harm to retain at least ipython here.
* Niggle: supported operating systems sentence can be updated to June 2022 in installation.rst as well (repeated from index.rst)

Working on checking everything works as expected with test script, then I see this as good to go.

TomWinder:
Slight curveball: I also tried installing from source on macOS, and have different behaviour compared to the previous setup.py :

```diff
- gcc -Wno-unused-result -Wsign-compare -Wunreachable-code -DNDEBUG -g -fwrapv -O3 -Wall -Wstrict-prototypes -I/Users/tom/miniconda3/envs/pyqt_faff/include -arch x86_64 -I/Users/tom/miniconda3/envs/pyqt_faff/include -arch x86_64 -I/Users/tom/Software/QuakeMigrate/quakemigrate/core/src -I/Users/tom/miniconda3/envs/pyqt_faff/lib/python3.7/site-packages/numpy/core/include -I/Users/tom/miniconda3/envs/pyqt_faff/include -I/Users/tom/miniconda3/envs/pyqt_faff/include/python3.7m -c quakemigrate/core/src/quakemigrate.c -o build/temp.macosx-10.9-x86_64-cpython-37/quakemigrate/core/src/quakemigrate.o -fopenmp -fPIC -Ofast
- creating build/lib.macosx-10.9-x86_64-cpython-37/quakemigrate/core/src
- gcc -bundle -undefined dynamic_lookup -L/Users/tom/miniconda3/envs/pyqt_faff/lib -arch x86_64 -L/Users/tom/miniconda3/envs/pyqt_faff/lib -arch x86_64 build/temp.macosx-10.9-x86_64-cpython-37/quakemigrate/core/src/quakemigrate.o -L/Users/tom/miniconda3/envs/pyqt_faff/lib -o build/lib.macosx-10.9-x86_64-cpython-37/quakemigrate/core/src/qmlib.cpython-37m-darwin.so -lm -lgomp

+ gcc -Wno-unused-result -Wsign-compare -Wunreachable-code -DNDEBUG -g -fwrapv -O3 -Wall -Wstrict-prototypes -I/Users/tom/miniconda3/envs/pyqt_faff/include -arch x86_64 -I/Users/tom/miniconda3/envs/pyqt_faff/include -arch x86_64 -I/Users/tom/Software/QuakeMigrate/quakemigrate/core/src -I/Users/tom/miniconda3/envs/pyqt_faff/include -I/Users/tom/miniconda3/envs/pyqt_faff/lib/python3.7/site-packages/numpy/core/include -I/usr/local/lib -I/Users/tom/miniconda3/envs/pyqt_faff/include/python3.7m -c quakemigrate/core/src/quakemigrate.c -o build/temp.macosx-10.9-x86_64-cpython-37/quakemigrate/core/src/quakemigrate.o -Xpreprocessor -fopenmp -fPIC -Ofast
+ creating build/lib.macosx-10.9-x86_64-cpython-37/quakemigrate/core/src
+ gcc -bundle -undefined dynamic_lookup -L/Users/tom/miniconda3/envs/pyqt_faff/lib -arch x86_64 -L/Users/tom/miniconda3/envs/pyqt_faff/lib -arch x86_64 build/temp.macosx-10.9-x86_64-cpython-37/quakemigrate/core/src/quakemigrate.o -L/Users/tom/miniconda3/envs/pyqt_faff/lib -L/usr/local/lib -o build/lib.macosx-10.9-x86_64-cpython-37/quakemigrate/core/src/qmlib.cpython-37m-darwin.so -lm -lomp
```

Net result is that it now does not build/install successfully. I think it is ultimately due to the change from `-lgomp` to `-lomp`, as lgomp is the only openMP library I have installed currently. Not sure if this and/or the other changes are expected from the revamped setup.py

hemmelig:
Can you run the following in iPython:

```
import platform
print(platform.system())
```

TomWinder:
Testing shows that the changed behaviour is due to a different (seemingly now more accurate) platform test in the updated `setup.py` -- looking now at the output of `platform.system()` rather than `pkg_resources.get_build_platform()`. Previously my macos system was using the default Unix flags, rather than the mac-specific ones, hence working only with `gcc`.

Whether to use `-lomp` etc. vs the standard Linux `-lgomp` etc. compiler / linker flags then depends whether a macos user is compiling with `clang` or `gcc` - both of which require manual install, particularly to satisfy the openMP dependency. A (not very thorough) comparison on Intel Macbook Pro / macos 10.15 Catalina suggests substantial performance benefits to using `clang`, so for now we'll take that as a reason to use this route: assuming by default that `clang` is being used on a macos system.

Other developers have faced the same issue - where it is difficult to distinguish whether a user is compiling with `gcc` vs `clang` (due to symlinking etc etc.) - and ultimately we haven't seen an elegant solution, so will put this on the wishlist pending time to burn fiddling with this. Hopefully with the move to pip it will become a non-issue in the near future in any case.

We will add a snippet to the docs outlining how this issue might present (if a macos user's default compiler is `gcc`), and how it can be circumvented (by temporarily setting the default compiler to `clang` - including installing the required dependencies with `brew` if required).

hemmelig:
I propose we get this merged. Current state of main branch is a little broken and I think the pinned package versions are catching up with us in terms of compatibility.

Need to bump version to 1.0.1 with this merge.

TomWinder:
I agree - I think there is just one thing outstanding, which is the final update of installation instructions in the docs. See attached for current state of changes on my branch. Will try finish this up urgently (need to add findings of the experiments I ran when we were at ROB).

[installation-docs_changes_20230302.docx](https://github.com/QuakeMigrate/QuakeMigrate/files/10872755/installation-docs_changes_20230302.docx)

TomWinder:
We also need to bump skfmm to 2022.08.15 (see https://github.com/TomWinder/QuakeMigrate/commit/48a4ca8b7e8e68a059cbff014266f338e410773e)

hemmelig:
There's a couple more places where `pyproj` needs updating (`create_lut.py`) to match up with the move away from `Proj`. I've got fixes on my working branch, will move them over.

TomWinder:
Kk I've addressed the instances of ``pyproj.transform()``; wasn't aware that ``pyproj.Proj()`` is also deprecated (there aren't any messages about this?).

TomWinder (commented):
Review of changes / original scope / etc.:
- wheel building is now **working** for all but macos-arm64; but there is an acceptable failover with pip install building from source on the user's machine (tested by @hemmelig). This satisfies the central goal of this PR
- version pins have been removed, which deals with a bunch of burgeoning issues related to deprecations and removal of support for e.g. old matplotlib versions. Also addresses https://github.com/QuakeMigrate/QuakeMigrate/issues/148
- supported Python versions updated to bring us up-to-date (3.8 - 3.11)
- wholesale review of ``installation.rst`` to improve compiler FAQ (and hopefully centralise our knowledge on this issue), and to prioritise installation via `pip`

Outstanding issues:
- ``Volcanotectonic_Iceland`` (dike example) test still **FAILS** - though this is consistent with the behaviour from before this PR. However I think this needs to be (urgently) dealt with in a separate PR.
- couple of very minor things which I will momentarily tweak and push, but otherwise I think we are good to merge. (🥳🥳🥳)

TomWinder (dismissed):

TomWinder (dismissed):

QuakeMigrate-developers (approved):
No consequential changes made since @TomWinder 's review. Documentation changes tested locally and README.md file previewed. Console behaviour not the same as displayed by VSCode markdown extension—text does not wrap—so may want to find a better solution.

GitHub Action fails at final step, but this is due to a prior test using the 1.0.1 version number (must have been last year). All wheels actually upload; it is only the .tar.gz that doesn't make it. This should not be an issue when the Action runs for the release and publication on PyPI for real, since the TestPyPI upload stage will be skipped (not a tag being pushed). Approving and moving to merge, then draft the release. Will hold off from finalising release until tomorrow.

hemmelig:
I'm going to leave this open for now until the release is finalised.

Issue #144, opened 2022-12-08
[BUG]: skfmm LUT generation doesn't work for the southern hemisphere


TomWinder:
**Describe the bug**
Attempted to use the "1dfmm" method to calculate an LUT for a grid in the southern hemisphere (Java) - traveltimes are clearly wrong. Suspect they may be mirrored along one axis; y seems most likely?

**To Reproduce**
Specify a grid that is in the southern hemisphere (ie a pair of negative latitudes - note the grid in question was also in the eastern hemisphere; not yet tested if there would be any / different problems in south&west), and calculate traveltimes using the "1dfmm" method. Manual inspection of traveltimes from various points in the LUT shows obviously incorrect results (e.g. the most distant station sometimes have the shortest traveltimes).

**Expected behavior**
Traveltimes increasing with distance from the chosen location. Note that several tests were performed to isolate this to the traveltime generation step, including:
 - using the same gridspec, station file etc. but with 'homogeneous' velocity model --> traveltimes increase with distance
 - inspection of grid coordinates of stations (i.e. do they correspond to expected locations)
 - plotting of velocity model to ensure it wasn't messed up somehow

**Desktop (please complete the following information):**
 - Operating System: Windows 11
 - Python version: 3.7
 - QuakeMigrate version: 1.0.0

**Additional context**
*NOT* yet compared to the "1dnlloc" traveltime generation. However, this method _has_ been used before in the same region with no such difficulties. This indicates it is likely isolated to the "1dfmm" implementation.

TomWinder:
Confirming the same bug affects Linux (Ubuntu 20.04 / Python 3.8 / QuakeMigrate 1.0.0):

`1dnlloc` traveltimes:
```
CBJI 4.93516939247
CBJM 7.46098649797
CCJM 6.92342233406
CWJM 7.81328655045
DBJI 9.79183836983
JPJI 8.55424903482
JTJM 10.6687429621
LEM 11.5750472581
PBJI 11.5753167666
SCJM 3.84283267304
SPSJM 11.4048941522
SSJM 5.62294754896
WSJM 10.8150932759
```

vs `1dfmm` traveltimes:

```
CBJI 4.74641187468
CBJM 5.55452430853
CCJM 6.64352053357
CWJM 2.92746251661
DBJI 5.3082381857
JPJI 3.32778470283
JTJM 8.40255334532
LEM 3.72185884826
PBJI 7.54072961752
SCJM 3.72839355312
[...]

Issue #137, opened 2022-03-02
[BUG]: All traces rejected where one pair can't be merged due to "Incompatible traces" error.


TomWinder:
**Describe the bug**
When the time period of data to be analysed contains a single pair of traces that can't be merged, but have overlapping/adjacent data & identical SEED id's, the whole merge operation is abandoned. This means all data over the day (or hour, etc.) boundary is not merged, leading to zero stations being available for migration.

**To Reproduce**
Include overlapping data that can't be merged in a way that results in this error (e.g. one file with "FLOAT64" encoding and the other with "STEIM1") in the data archive.

**Expected (desired) behaviour**
Only the station/stream which is affected by this merge error should be rejected.

**Desktop (please complete the following information):**
 - Operating System: Ubuntu 16.04
 - Python version: 3.8.4
 - QuakeMigrate version: 1.0.0

**Additional context**
This is annoying to fix because of the way ObsPy's `st.merge()` handles the `method=-1` option; in that case it does the operation inplace only, and doesn't explicitly return the `Stream` object (preventing e.g. application of the merge operation on a station-by-station basis). I will raise a linked issue there. For the time being could do:

```
ids = set([tr.id for tr in st])
st_merged = Stream()
for tr_id in ids:
    st_merged += st.select(id=tr_id)._cleanup()
```

(i.e. directly using the hidden `st._cleanup()` function).

Issue #134, opened 2021-10-31
BUG: Add "units" arg to to_mfast.py


amberparsons1:
Currently to_mfast.py assumes QM outputs event depths and uncertainties in metres but this has now changed to km.

Add a 'units' arg (see to_obspy.py).

PR #151, opened 2023-04-09
FIX: more minor bug fixes addressing the list of current Issues

TomWinder:
To address a couple of minor bugs highlighted by recently submitted Issues.

### Relevant Issues
#144 
#137 
#134 

## Test cases
 - skfmm error tested with a modified dike_intrusion_lut where the grid was first left as-is ie not including all stations; then modified so that it did include all stations. All worked as intended.
 - new merge strategy tested by adding a second copy of `FLUR_E.m` for the `Volcanotectonic_Iceland` example, with encoding changed to `FLOAT64`, meaning it would suffer from the failed merged scenario encountered in #137 . All worked as intended: error message highlighting the errant channel, but migration continued successfully using all unaffected data.
 - to_mfast checked by applying this to the outputs from the `Volcanotectonic_Iceland` example. @hemmelig if you could double-check this with the eyes of an MFAST user that would be appreciated plz!

- [x] test_benchmarks.py reports all example tests pass  **except known error on ``Volcanotectonic_Iceland`` example**

### System details
- macOS 11.4 Big Sur (x86_64)
- Python 3.9

### Final checklist
- [x] `master` base branch selected?
- [x] All tests still pass.
- [x] First time contributors have added your name to `CONTRIBUTORS.md`.

### DIFF
--- a/quakemigrate/io/data.py
+++ b/quakemigrate/io/data.py
@@ -309,9 +309,9 @@ class Archive:
                     try:
                         st_merged += st.select(id=tr_id).merge(method=-1)
                     except UserWarning as e:
-                        print(f"\t\t{e}")
-                        print(f"\t\t{st.select(id=tr_id)}")
-                        print(
+                        logging.info(f"\t\t{e}")
+                        logging.info(f"\t\t{st.select(id=tr_id)}")
+                        logging.info(
                             "\t\tThis channel will not be used for onset "
                             "calculation."
                         )
--- a/quakemigrate/io/data.py
+++ b/quakemigrate/io/data.py
@@ -297,24 +297,8 @@ class Archive:
                     logging.info(f"File not compatible with ObsPy - {file}")
                     continue
 
-            # Merge all traces with contiguous data, or overlapping data which
-            # exactly matches (== st._cleanup(); i.e. no clobber)
-            # Apply this on a channel by channel basis so that if any individual
-            # merge fails then only that channel will be omitted.
-            ids = set([tr.id for tr in st])
-            st_merged = Stream()
-            with warnings.catch_warnings():
-                warnings.filterwarnings("error")
-                for tr_id in ids:
-                    try:
-                        st_merged += st.select(id=tr_id).merge(method=-1)
-                    except UserWarning as e:
-                        logging.info(f"\t\t{e}")
-                        logging.info(f"\t\t{st.select(id=tr_id)}")
-                        logging.info(
-                            "\t\tThis channel will not be used for onset "
-                            "calculation."
-                        )
+            # Merge waveforms channel-by-channel with no-clobber merge
+            st = util.merge_stream(st)
 
             # Make copy of raw waveforms to output if requested
             data.raw_waveforms = st.copy()
--- a/quakemigrate/io/data.py
+++ b/quakemigrate/io/data.py
@@ -123,15 +123,17 @@ class Archive:
         # Response removal parameters
         self.response_inv = kwargs.get("response_inv")
         response_removal_params = kwargs.get("response_removal_params", {})
-        if self.response_inv and \
-            "water_level" not in response_removal_params.keys():
-            msg = ("Warning: 'water level' for instrument correction not "
-                    "specified. Set to default: 60")
+        if self.response_inv and "water_level" not in response_removal_params.keys():
+            msg = (
+                "Warning: 'water level' for instrument correction not "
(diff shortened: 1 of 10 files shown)


quakemigrate/io/data.py
```diff
@@ -663,18 +720,22 @@ def get_real_waveform(self, tr, velocity=True):
...
-                            sacsim=True,  # To replicate remove_response()
-                            pitsasim=False)  # To replicate remove_response()
+                tr.simulate(
+                    paz_remove=paz_dict,
+                    pre_filt=self.pre_filt,
+                    water_level=self.water_level,
+                    taper=True,
+                    sacsim=True,  # To replicate remove_response()
+                    pitsasim=False,
+                )  # To replicate remove_response()
```
hemmelig, line 738:
```suggestion
                    pitsasim=False,   # To replicate remove_response()
                )
```

quakemigrate/io/data.py
```diff
@@ -722,18 +785,20 @@ def get_wa_waveform(self, tr, velocity=False):
...
-                    sacsim=True,  # To replicate remove_response()
-                    pitsasim=False)  # To replicate remove_response()
+        tr.simulate(
+            paz_simulate=util.wa_response(obspy_def=True),
+            pre_filt=self.pre_filt,
+            water_level=self.water_level,
+            taper=True,
+            sacsim=True,  # To replicate remove_response()
+            pitsasim=False,
+        )  # To replicate remove_response()
```
hemmelig, line 801:
```suggestion
            pitsasim=False,  # To replicate remove_response()
        )
```

quakemigrate/io/data.py
```diff
@@ -271,15 +287,34 @@ def read_waveform_data(self, starttime, endtime, pre_pad=0., post_pad=0.):
...
+                for tr_id in ids:
+                    try:
+                        st_merged += st.select(id=tr_id).merge(method=-1)
+                    except UserWarning as e:
+                        print(f"\t\t{e}")
+                        print(f"\t\t{st.select(id=tr_id)}")
+                        print(
+                            "\t\tThis channel will not be used for onset "
+                            "calculation."
+                        )
```
hemmelig, line 317: Is there a reason why these are `print` statements, and not logging?

hemmelig (commented):
Again, all minor changes. Some autoformatting applied (using black), which makes sense as the files were already being touched. A package-wide application of black will follow in a future PR (as this will touch the entire package). It shouldn't stall this PR, but perhaps you might add a test demonstrating the issue addressed by 065c587—have the test catch the user warning? Also, just a question about why `print` is being used here over `logging`?

It's still useful to have the `to_mfast` utility, but will ultimately be superseded by [AnisotropIO](https://github.com/hemmelig/AnisotropIO/). The unit keyword should be carried over to there, but I think the eventual goal here is to use this external package as an optional dependency (maybe adding an install mode like `pip install quakemigrate[utils]`).

Approving.

quakemigrate/io/data.py
TomWinder: Yeh I also was surprised by this, but it's because it's effectively a standalone function (rather than being called within something that's covered by a QM Run object), so there's no logger spun up to be served these messages

TomWinder: No, sorry, I'm being an imbecile. This is not where I thought it was.. fix incoming. I'll try to add a test, too.

hemmelig (approved):
Comments addressed. Go ahead with merge.

TomWinder:
Sorry for faff -- no rush. Appropriately in the process of writing a test, I discovered that my implementation of the channel-by-channel merge was in fact subtly incorrect (covered up by classic Python list faff).

quakemigrate/util.py
```diff
@@ -592,6 +593,44 @@ def upsample(trace, upfactor, starttime, endtime):
 
     return out
 
+def merge_stream(stream):
+    """
+    Merge all traces with contiguous data, or overlapping data which
+    exactly matches (== st._cleanup(); i.e. no clobber). Apply this on a
+    channel by channel basis so that if any individua merge fails then only
```
hemmelig, line 601:
```suggestion
    channel by channel basis so that if any individual merge fails then only
```

hemmelig (dismissed):
Great—I guess sometimes even just writing the test is valuable!

Approving and will move to merge.

hemmelig (approved):

PR #157, opened 2023-04-19
Add GitHub workflow for testing/code coverage assessment

hemmelig:
This PR implements a GitHub Action the runs the QuakeMigrate examples, assesses the code coverage these examples provide, and runs the benchmark comparison test. The Action is triggered by any of the following:

- Pushes to branches prefixed with:
  - `feature/`
  - `bugfix/`
  - `patch/`
  - `tests/`
- Pull requests where the target branch is either:
  - `master`
  - or is prefixed with `version/`
- The publication of a release

Tests are run for Python versions 3.8, 3.9, 3.10, and 3.11. Currently these tests are only run on the latest Ubuntu operating system—it would likely be wise to extend this to at least macOS, if not Windows as well (though we would have to find a solution to the NonLinLoc install issue with Windows. I have not yet tried to install NonLinLoc via the Cmake build system now available via the GitHub-hosted version—worth a try.

### System details
This update is run on the GitHub Actions system. Tested across multiple Python versions and Ubuntu (latest = 22.04)

### Final checklist
- [x] Correct base branch selected? - This change sits explicitly on top of the `patch/v1.0.2` branch.
- [x] All tests still pass.
- [x] First time contributors have added your name to `CONTRIBUTORS.md`.

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -41,26 +41,29 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install '.[dev]'
-      
+
+      # - name: Collect NonLinLoc sourcecode
+      #   uses: actions/checkout@v3
+      #   with:
+      #     fetch-depth: 0
+      #     path: NonLinLoc
+
       # - name: Install NonLinLoc
+      #   working-directory: ./NonLinLoc
       #   run: |
-      #     mkdir nonlinloc
-      #     cd nonlinloc
-      #     curl http://alomax.free.fr/nlloc/soft7.00/tar/NLL7.00_src.tgz -o NLL7.00_src.tgz
-      #     tar -xzvf NLL7.00_src.tgz
       #     cd src
       #     mkdir bin
-      #     make -R all
-      #     mv Vel2Grid Grid2Time bin/.
+      #     cmake .
+      #     make
       #     echo "$PWD/bin" >> $GITHUB_PATH
 
       - name: Run Iceland icequake example
         working-directory: ./examples/Icequake_Iceland
         run: |
-          coverage run --source=quakemigrate --data-file=../../.coverage -a iceland_lut.py 
+          coverage run --source=quakemigrate --data-file=../../.coverage -a iceland_lut.py
           coverage run --source=quakemigrate --data-file=../../.coverage -a iceland_detect.py
           coverage run --source=quakemigrate --data-file=../../.coverage -a iceland_trigger.py
-          coverage run --source=quakemigrate--data-file=../../.coverage  -a iceland_locate.py
+          coverage run --source=quakemigrate --data-file=../../.coverage -a iceland_locate.py
 
       # - name: Run Iceland volcano-tectonic example
       #   working-directory: ./examples/Volcanotectonic_Iceland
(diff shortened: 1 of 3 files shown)


hemmelig:
Will squash into a single commit on merge.

hemmelig:
The prognosis for running tests (and QuakeMigrate in general) on Windows where NonLinLoc is desired is...not good! Thankfully, it remains possible to run QM on Windows using either the other options for traveltime lookup table generation OR it should be possible to build the LUTs on a *nix system with NonLinLoc, then transfer them over to a Windows machine and proceed from there.

For testing purposes on Windows, it might be worth taking this latter approach. Construct the Windows-compatible LUT file, put it somewhere that can be grabbed (could place in the repository under the example), and adjust the Windows runners to load the LUT this way, rather than generate it fresh. However, _I do not think this should delay this merge._

TomWinder (approved):
Overall looks great. A couple of queries, but nothing show-stopping.

1. What was the issue that prevented using the NonLinLoc install from tar? It probably is sensible to switch over to the GitHub source, but if so then we should also update the installation instructions etc so that it will be consistent when people run the tests locally after install.
2. Is there a way to run tests at some set interval (e.g. weekly?) in the same way we used to with Travis?
3. Any reason not to also add `test_import.py` to the list?
4. Any reason not to use `run_test_examples.py` as opposed to running each script individually in the current workflow? I can see it has to do with the coverage stuff, but surely e.g. `test_util.py` should also contribute to the coverage stats..

hemmelig:
> 1. What was the issue that prevented using the NonLinLoc install from tar? It probably is sensible to switch over to the GitHub source, but if so then we should also update the installation instructions etc so that it will be consistent when people run the tests locally after install.

Unsure—the [output from the workflow](https://github.com/QuakeMigrate/QuakeMigrate/actions/runs/4739283027/jobs/8413964003#step:5:1725) had this to say when trying to build from the `.tar.gz` file:

```
collect2: error: ld returned 1 exit status
make: *** [Makefile:100: NLLoc] Error 1
Error: Process completed with exit code 2.
```

> 2. Is there a way to run tests at some set interval (e.g. weekly?) in the same way we used to with Travis?

Yes—add a schedule section to the `on` header of the workflow file, e.g. run tests every Sunday at 3:30 am:

```
  on:
    schedule:
    - cron: 30 3 * * 0
```

> 3. Any reason not to also add test_import.py to the list?

Arguably, yes—this is not a coverage test, more of a failure test. The tests ought to fail for the same reason the `test_imports.py` would fail, and then we would just need to either run that test ourself or look at the output to identify the culprit. Might be worth having a distinct section for each architecture where the tests are only run as tests and a separate one for coverage?

> 4. Any reason not to use run_test_examples.py as opposed to running each script individually in the current workflow? I can see it has to do with the coverage stuff, but surely e.g. test_util.py should also contribute to the coverage stats..

I guess not—I was copying across from the state of the TravisCI testing config, which had each coverage line ran independently.

.github/workflows/run_tests.yml
```diff
@@ -0,0 +1,110 @@
+name: Run tests and upload code coverage report
+on:
+  push:
```
hemmelig, line 3:
```suggestion
  schedule:
  - cron:
    30 3 * * 0
  push:
```

TomWinder:
Happy for this to be merged. Let's just add a note somewhere about finding a streamlined way to add (particularly) unit tests to the tests job script.

Great to get CI up and running again 👍

Issue #136, opened 2022-02-24
ImportError: Could not load extension library 'qmlib'.


glPeng-seis:
I installed quakemigrate according to the steps of the instructions. But I still meet a problem, like this:
===================================
```
>>> import quakemigrate
Traceback (most recent call last):
  File "/Users/guanlingpeng/QuakeMigrate/quakemigrate/core/libnames.py", line 39, in _load_cdll
    cdll = ctypes.CDLL(str(lib))
  File "/Users/guanlingpeng/anaconda3/envs/quakemigrate/lib/python3.7/ctypes/__init__.py", line 364, in __init__
    self._handle = _dlopen(self._name, mode)
OSError: dlopen(/Users/guanlingpeng/QuakeMigrate/quakemigrate/core/src/qmlib.cpython-37m-darwin.so, 0x0006): tried: '/Users/guanlingpeng/QuakeMigrate/quakemigrate/core/src/qmlib.cpython-37m-darwin.so' (no such file), '/usr/local/lib/qmlib.cpython-37m-darwin.so' (no such file), '/usr/lib/qmlib.cpython-37m-darwin.so' (no such file)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/guanlingpeng/QuakeMigrate/quakemigrate/__init__.py", line 18, in <module>
    from quakemigrate.signal import QuakeScan, Trigger  # NOQA
  File "/Users/guanlingpeng/QuakeMigrate/quakemigrate/signal/__init__.py", line 20, in <module>
    from .scan import QuakeScan  # NOQA
  File "/Users/guanlingpeng/QuakeMigrate/quakemigrate/signal/scan.py", line 23, in <module>
    from quakemigrate.core import find_max_coa, migrate
  File "/Users/guanlingpeng/QuakeMigrate/quakemigrate/core/__init__.py", line 20, in <module>
    from .lib import migrate, find_max_coa  # NOQA
  File "/Users/guanlingpeng/QuakeMigrate/quakemigrate/core/lib.py", line 20, in <module>
    qmlib = _load_cdll("qmlib")
  File "/Users/guanlingpeng/QuakeMigrate/quakemigrate/core/libnames.py", line 45, in _load_cdll
    raise ImportError(msg)
ImportError: Could not load extension library 'qmlib'.

[...]

PR #161, opened 2023-09-29
Sorted issue with axis sharing by commenting out offending lines.

TomSHudson:
Sort issue with joining axes in order to share them in various submodules. Issue solved by commenting these lines (see ###!!!) without change in core functionality. 

### Relevant Issues
n/a.

## Test cases
Run Icequake_Iceland example. 
Note: test_benchmarks.py has other issues: "ImportError: Could not load extension library 'qmlib'." so didn't run tests.

### System details
Mac OS 10.14.6 but changes shouldn't have any effect on other OSs.

### Final checklist
- [ Y] Correct base branch selected?
- [ N] All tests still pass.
- [ N] Any new features or fixed regressions are covered by new tests.
- [ Y] Any new or changed features are fully documented.
- [ Y] Significant changes have been added to `CHANGES.md`.
- [ Y] First time contributors have added your name to `CONTRIBUTORS.md`.

### DIFF
--- a/quakemigrate/lut/lut.py
+++ b/quakemigrate/lut/lut.py
@@ -674,8 +674,8 @@ class LUT(Grid3D):
         xz = plt.subplot2grid(gs, (7, 0), colspan=5, rowspan=2, fig=fig)
         yz = plt.subplot2grid(gs, (2, 5), colspan=2, rowspan=5, fig=fig)
 
-        xz.get_shared_x_axes().join(xy, xz)
-        yz.get_shared_y_axes().join(xy, yz)
+        ###!!!xz.get_shared_x_axes().join(xy, xz)
+        ###!!!yz.get_shared_y_axes().join(xy, yz)
 
         # --- Set aspect ratio ---
         # Aspect is defined such that a circle will be stretched so that its
--- a/quakemigrate/plot/phase_picks.py
+++ b/quakemigrate/plot/phase_picks.py
@@ -62,7 +62,7 @@ def pick_summary(event, station, waveforms, picks, onsets, ttimes, windows):
     axes = fig.axes
 
     # Share P-pick x-axes and set title
-    axes[0].get_shared_x_axes().join(axes[0], axes[3])
+    ###!!!axes[0].get_shared_x_axes().join(axes[0], axes[3])
     axes[0].set_xticklabels([])
     axes[0].set_yticklabels([])
     axes[0].yaxis.set_ticks_position("none")
@@ -71,7 +71,7 @@ def pick_summary(event, station, waveforms, picks, onsets, ttimes, windows):
 
     # Share S-pick x-axes and set title
     for ax in axes[1:3]:
-        ax.get_shared_x_axes().join(ax, axes[4])
+        ###!!!ax.get_shared_x_axes().join(ax, axes[4])
         ax.set_xticklabels([])
         ax.set_yticklabels([])
         ax.yaxis.set_ticks_position("none")
--- a/quakemigrate/plot/trigger.py
+++ b/quakemigrate/plot/trigger.py
@@ -109,8 +109,8 @@ def trigger_summary(
         ax.set_xlim([starttime.datetime, endtime.datetime])
 
     # --- Plot LUT, coalescence traces, and station availability ---
-    for ax in fig.axes[:2]:
-        ax.get_shared_x_axes().join(ax, fig.axes[2])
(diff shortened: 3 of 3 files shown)


TomSHudson:
Some tests still aren't working, but tests get further than before.

TomWinder:
Hi Tom, we're aware of this issue. The best workaround for the time being is to pin matplotlib to 3.7:

In environment.yml and .binder/environment.yml:
`matplotlib = 3.7`
And in requirements.txt:
`matplotlib==3.7`

We are not going to merge these changes into master  -- commenting those lines skips the bug but will also mean the plots are not produced properly. We've noticed some strange behaviour with test failures recently, and @hemmelig has investigated but we don't yet fully understand the cause (working locally, but not on the CI server..) -- so running the tests locally and/or checking the binder notebooks is a sensible for now.

Your error ``"ImportError: Could not load extension library 'qmlib'"`` indicates that the new version has not been properly installed in the environment you are running in. Worth double-checking the paths etc., and submit an issue with more details if it persists.

As a separate note, if you're planning to use binder to demonstrate stuff, it's currently pretty crippled due to Google having removed support - suggest using the workaround here: https://github.com/jupyterhub/binderhub/issues/1749 (i.e. using a link that directs specifically to a GESIS node.), and they also have an issue building environments with Python 3.11. Suggest pulling from https://github.com/TomWinder/QuakeMigrate/tree/prac_v1 , which has workarounds for this that are functional as of this morning. E.g. link is here: https://notebooks.gesis.org/binder/v2/gh/TomWinder/QuakeMigrate/prac_v1?urlpath=lab%2Ftree%2Fexamples%2FVolcanotectonic_Iceland_prac%2FQM_prac.ipynb

hemmelig:
Where are you running the tests from locally? Did you install the environment from PyPI or build locally from source? What platform are you running the code on?

The `qmlib not found` issue often occurs when running in the base git directory i.e. the local path to the `quakemigrate` module takes precedence over the path to the one in whatever environment you're working in (e.g. `conda`). Given that you seem to be able to run the examples—that is, `qmlib` is being found—I suspect that's the issue?

In terms of the PR—just commenting out the lines isn't something that should be merged to master. If they're no longer needed, we can remove them outright, otherwise I believe @TomWinder made a fix in the last couple of days on his local branch.

The tests being run on the GitHub workflow machines are failing for an unknown reason—we haven't been able to replicate the issue on any of the platforms we have access to (macOS + Ubuntu 22.04 linux). When you look at the workflow log, the issue is (I believe) in a PickTime, differing by one digit at the 6th decimal place, on one line. So some very obscure floating point difference is cropping up somewhere. I believe there are ways to access the machines that run the tests, which is likely the best way to diagnose exactly what is happening, but haven't yet decided what to do about it. I searched through all of the potential changes to packages, but like I said, I could not replicate the error locally.

TomSHudson:
Sure ok, thanks.

I only wanted to fix the issues promptly as master branch should never have any issues. Let me know if you'd like help fixing the issues raised by the tests, as it would be good to get master running again without issues ASAP.

I had the same issues whether I installed via PyPi or a git clone. 

Maybe as a quick fix for now, I'd suggest pinning that matplotlib dependency not only in the requirements but also ensuring that the pip build also requires it.

FYI, fixing that issue for me only led to more tests failing (e.g. Volcanotectonic_Iceland).

Thanks for the quick response. I'll close this pull request.

TomWinder:
Yes, it was only discovered yesterday. I'll submit an Issue describing it properly and we'll test and push the version pin fix as soon as we can. If you could help by testing this change locally and submitting a new PR that would help us get master fixed as quickly as possible.

Could you clarify whether the "more tests failing" is the automatic tests running on Github (see details about this from Conor), or something failing locally? If it's failing locally then that's something new which we would investigate.

If it's the same issue with pip install vs install from Git source then what Conor mentions seems most likely (attempting to import quakemigrate from the "QuakeMigrate" source dir - where it looks first at the 'quakemigrate' dir in your working directory, rather than the proper install path). See here: https://github.com/QuakeMigrate/QuakeMigrate/issues/136

TomWinder:
Great - see comments on issue #162 for my suggestion.

Yes we'll prioritise getting the tests running again reliably - hadn't looked until now but you're right the matplotlib issue had taken over as the cause of failures since 2 weeks ago (since matplotlib 3.8 release - makes sens), and we hadn't noticed due to the ongoing floating-point weirdness making the tests fail anyway.

Issue #172, opened 2024-03-07
Tests timing out on data acquisition from EarthScope stage


hemmelig:
**Describe the bug**
Not strictly a bug, but the tests are failing for, I believe, reasons of being timed out by the EarthScope DMC. They've recently moved to using a cloud platform for their services, so this could be a protection measure that stops them from being swamped with excessive requests? We might consider moving the example waveform files to a dedicated Zenodo and using curl/wget to retrieve them, then unpack them.

**To Reproduce**
Based on the reports from the last automated test run.

**Expected behavior**
The tests should run without issue.

TomWinder:
If it's definitely an anti-spam problem (seems likely, though couldn't immediately see any bug reports about this on obspy etc), then I think there are ~ 3 options:
1. Zenodo, as suggested
2. Look into using the [mass_downloader](https://docs.obspy.org/packages/autogen/obspy.clients.fdsn.mass_downloader.html#module-obspy.clients.fdsn.mass_downloader) -- but for this particular example I think this is probably overkill. I was sure I'd written a neat script using this for the forthcoming Askja example, but I can't find it right now..
3. edit the download script to just make one request, with all stations (checked this manually and it's also way faster):

```
station_string = ",".join(stations["Name"])
```

-->
```diff
@@ -29,17 +29,16 @@ endtime = UTCDateTime("2014-236T00:15:00")

#  --- Read in station file ---
 stations = read_stations(station_file)
+stations_string = ",".join(stations["Name"])
 
 # --- Download instrument response inventory ---
-inv = Inventory()
-for station in stations["Name"]:
-    inv += client.get_stations(
-        network=network,
-        station=station,
-        starttime=starttime,
-        endtime=endtime,
-        level="response",
-    )
+inv = client.get_stations(
+    network=network,
+    station=stations_string,
+    starttime=starttime,
+    endtime=endtime,
+    level="response",
+)
 inv.write(response_file, format="STATIONXML")
 
 # --- Make directories to store waveform data ---
@@ -47,20 +46,22 @@ waveform_path = data_path / str(starttime.year) / f"{starttime.julday:03d}"
 waveform_path.mkdir(parents=True, exist_ok=True)
 
 # --- Download waveform data ---
+print(f"Downloading waveform data from {datacentre}")
+st = client.get_waveforms(
+    network=network,
+    station=stations_string,
+    location="*",
+    channel="*H*",
+    starttime=starttime,
+    endtime=endtime,
+)
[...]

PR #174, opened 2024-03-07
:bug: Single client request for all stations

TomWinder:
Previously data was downloaded from each station sequentially in a loop; this instead makes use of the ability to list multiple stations in a single request. This should speed things up a little, and hopefully address #172

Initiated by failure of the automatic test run, seemingly due to timing out during data download for the VT-Iceland example. Hopefully by reducing requests we stay beneath any traffic limits, but this also provides a significant speed-up in the data download.

### Relevant Issues
#172 - [Tests timing out on data acquisition from EarthScope stage](https://github.com/QuakeMigrate/QuakeMigrate/issues/172)

## Test cases

### System details
Ubuntu 22.04

### Final checklist
- [x] All tests still pass.


hemmelig (approved):
Good to go.

PR #179, opened 2024-09-06
FEAT: Add trigger median_ratio threshold & smoothing

TomWinder:
Introduces a new method to determine the trigger threshold, by taking a multiplier of the median value of the coalescence trace, in a user-specified window duration.

So far the existing MAD multiplier method is not renamed from "dynamic"; instead the new method is identified by an additional `threshold_method` option "median_ratio". We might consider renaming from "dynamic" to "mad_ratio" or similar to avoid ambiguity, but this would be a breaking change.

Additional small tweaks surroundng this feature addition:
 - pass `threshold_method` through to the trigger_summary() plotting function
 - add an example usage of the median_ratio threshold in the Volcanotectonic_Iceland example trigger() script
 - fix a bug in `trigger.chunks2trace()` caused by the deprecation of `numpy.product()` (--> `np.prod()` )

Remaining to-do:
- [x] Add trigger smoothing
- [x] Consider renaming of existing "dynamic" trigger threshold method
- [x] Unit test(s) for trigger threshold methods
- [x] Unit test for trigger smoothing

### Relevant Issues
N/A

## Test cases

### System details
Ubuntu 22.04 LTS / Python 3.12.4 / AMD EPYC

### Final checklist
    - [x] rebase `version/v1.2.0` before merge (I think)
- [x] All tests still pass.
- ~Any new or changed features are fully documented.~


hemmelig:
We could likely soft deprecate dynamic (though that also raises the question of when we properly deprecate the things we've already added warnings for, would be nice to clean up!).

hemmelig:
Deprecation handling:

```python
    # --- Deprecation/Future handling ---
    @property
    def threshold_method(self):
        """Handle deprecated 'dynamic' threshold method."""
        return self._threshold_method

    @threshold_method.setter
    def threshold_method(self, value):
        """
        Handle deprecated threshold_method kwarg / attribute and print warning.

        """

        if value == "dynamic":
            # DEPRECATED
            print(
                "FutureWarning: This threshold method has been renamed - continuing.\n"
                "To remove this message, change:\n\t'dynamic' -> 'mad'"
            )
            self._threshold_method = "mad"
        else:
            self._threshold_method = value
```

And then change line 191 to:

```python
        elif self.threshold_method == "mad":
```

hemmelig:
Few additional changes to propagate `dynamic` -> `mad`, but I can add them as suggestions in a review, or push as a commit.

TomWinder:
Note also: tested the deprecation handling by running the VT_Iceland example with:
- static (and checking benchmarks);
- "dynamic" (FutureWarning appears correctly);
- "mad"; and
- "median_ratio"

All work as expected (and new trigger summary text renders as expected).

TomWinder:
Remaining quandary (aside from writing some tests #urgh) is the what to do about peak-index finding - in some previous testing (and usage) that I can't quite remember, I had some issues with the `COA_N` and `COA` maxima being more than a couple of samples apart, which led to problems in locate (which is locked to using `COA` to find the origin time of the event, and hence determining whether it falls within the marginal window buffer around the trigger time) sometimes rejecting "real" events. By fixing the peak to be identified from the `COA` trace, this issue was resolved.

Admittedly these events were outliers (M~5's in Bb), which is the real underlying cause for the issues experienced, but does at least highlight the potential impact of this choice.

I plan to go back to that example with the current code and do some experimenting, then will feed back my opinion here..

TomWinder:
Please take a look at the tests - I think there were a few options how to do it (in terms of instantiating a Trigger object or not, etc), but this seemed a safe way to do it.

Otherwise I think this just needs updates to CHANGES.md (within this PR, or at the point of merging to master from `version/v1.2.0`?) and then maybe some bits adding to docs (if trigger docs even really exist; if not, will skip.)

TomWinder:
The troublesome event:
image

So in re-inspection, this is purely an issue of "COA" vs "COA_N", and not affected by the smoothing (or threshold method); in both cases (smoothed and un-smoothed) the "COA" peak is close (enough) to the final origin time from locate (2017-03-08T03:09.88), while the COA_N trace peaks later, and far enough away from the peak in the locate COA timeseries that it is rejected as outside the marginal window (here 1 second). (Apologies that the ordering of traces in the plot is not very logical).

For context, the locate COA timeseries has a similar flat-topped shape, but mirroring the COA timeseries from detect() with the early peak being the highest:

image

I think it's quite tricky to know what the best general solution is for this. Theoretically/philosophically, I stand by the reasoning that the COA timeseries reflects what is used to determine the origin time in locate, and is therefore the most direct comparison (given the relatively strict criteria on origin time for the event to be "accepted"). However I can imagine scenarios where identifying peaks based on one timeseries but the index of the max value on another might cause unexpected behaviour. That said, the idxmax() -- in `trigger._identify_candidates()`; line 510 -- is heavily bounded by only being given the "candidate" data window to look at, so worst-case scenario it sets the "peak" as the first/last sample within the candidate window.

We did originally have this hard-coded to always refer to the "COA" timeseries, and it was reverted in 7a5389c .

hemmelig (commented):
All current changes look good to approve. Couple of minor notes:

1. I think we can drop the `-*- coding: utf-8 -*-`, on reading about it I think it's only strictly required for Python 2.7, as Python 3 uses UTF-8 by default.
2. I wrote an explicit test for the `chunks2trace` function that we could add to the `test_trigger.py` suite. Might help track down an issue slightly quicker and no real harm (wanted to have a review of writing tests and started there as didn't want to tackle the spoofing of the inputs for `Trigger`. For testing purposes we could use the `mock` package, instead of passing in strings. We might also consider having some sort of default `LUT` that has a couple of stations and uses a homogeneous velocity model somewhere either in the package or just for testing purposes. Not strictly necessary now.

hemmelig:
I think for the purpose of this PR, the peak-finding issue can be addressed separately (since it is not actually affected by the smoothing or threshold value).

Do you have a proposal in mind that would address the example scenario you presented above? In some ways, isn't the marginal window the final control on whether an event is kept? This kind of touches on something I noticed when working on the updated icequake examples—the width of the peak around the origin time for events can actually vary quite a lot, so using a single marginal window doesn't always seem appropriate. Making an estimate of the marginal window based on the width of the peak in the coalescence might be a way of addressing this (to an extent) in an automated/analytical way?

TomWinder:
Thanks for checking through it.

1. Yes can remove if preferred. I just copied `test_onsets.py` as a starting point, so all the formatting mirrors that.
2. For the purposes of this PR I think we should just leave as-is if it works. It exposes that there should probably be a check that the `LUT` passed to `Trigger` actually is an LUT object. I'd lean towards making our own mock LUT to use, but I don't think this is the time to start pulling that thread.

Re. peak finding:
 - the proposal is to reverse  the change made in 7a5389c : 
```diff
- peak = candidate.loc[candidate[trigger_on].idxmax()]
+ peak = candidate.loc[candidate["COA"].idxmax()]
```
 - re. the stability issues etc. I mentioned, with hindsight I've tested this on > 45 years' worth of runs across Askja/Vatnajökull/Reykjanes etc, and haven't come across any problem. And with this setting haven't come across a recurrence of the issue that is outlined (and I have checked similar events - ie M5's in Bb - specifically, though haven't cross-compared to the normalised coalescence trace).
 - having mulled this a bit more I think for now it is probably the right way to go. It could be left to a separate PR, but I'm not sure what the benefit of that is (though it definitely adds git admin)

Re. marginal window:
 - yes I've played with this previously as well. But I think this is a more difficult problem to solve than it maybe first seems from the ~ well-behaved icequake examples. Getting a reference for the peak height (in order to get the width) will give us similar issues to the 3D gaussian fitting for quite a few small events I've seen (baseline can be quite variable in comparison to absolute peak height, and non-symmetrical), and in terms of the event-distinction logic the marginal window would only ever be able to be shrunk, not expanded, else you might end up locating different events than you triggered on (à la CMM.......).
 - That said, I've also thought for a while that characterising the peak (i.e. by fitting a Gaussian, or similar) might be another useful way to distinguish "well-located" events, and that seems like the obvious starting point for this. Anyway, I think this is _mostly_ a separate issue of optimising the location (&location uncertainty estimate) in locate, rather than being the key player in this particular problem.

hemmelig:
Probably only worth removing them gradually whenever we touch a file for other reasons OR do it all in one blanket pass, but really no rush.

Yeah, taking the Gaussian fitting procedure from the picking stage was my thinking too. For the future though.

Ok—go ahead with that change. Once that's done, I think I'm happy to approve this PR and get it merged to version/v1.2.0. Is there anything else we want to do before re-running the paper examples with the latest version? I have a mostly worked up implementation of a command-line interface, but can wait for now. Perhaps approving some of the docs/synthetic example code would be good to go in v1.2.0?

TomWinder:
Great, will do that shortly.

I have been working on my remaining residual changes today:
1. Some minor bugs in amplitudes/magnitudes (one edge case that churning lots of data has exposed, and another bug that basically just affects the amplitudes plot). Will push this in 2 mins and open a separate PR aimed at master.
2. Built-in calculation of COV_Err_XYZ (geometric mean of the three COV_Err's) = 8 additional lines in io/event.py, but will also then change the benchmark locate outputs (by adding a column to .event files). Please lmk where you would prefer this to be pushed (can be within this PR, in its own (to which branch?)), whatever.

After that I'm raring to draw a line in the sand and re-do the examples. Only two things I can think of are:
1. Adding synthetic (+/- associated docs depending how ready-to-go they are) and Askja examples
2. Code modifications necessary to produce the figures from the above (main thing I can recall is saving the marginalised 3D coa_map for plotting -- I added an io.marginal_coalescence etc for this; can be pushed along with the example). Don't recall if you mentioned making any changes for running/plotting the synthetic.

TomWinder:
Also on my non-urgent list are CodeCov, and switching the Zenodo repository link on README.md. (Probably there are some others like this).

hemmelig:
I'll look into the CodeCov situation. I think it's likely I also needed to write out the coalescence for plotting the synthetic examples (will double check).

hemmelig (approved):
Approved to merge.

Issue #180, opened 2024-09-09
BUG: Crash when attempting to plot rejected amplitude measurements


TomWinder:
**Describe the bug**
The following error leads to a fatal crash when attempting to plot rejected amplitude measurements:

```
Traceback (most recent call last):
  File "/home/tebw2/QMigrate/RUN_SCRIPTS/TW_locate_node_RK_qmv1_1.py", line 166, in <module>
    scan.locate(starttime=start, endtime=end)
  File "/home/tebw2/Software/QuakeMigrate/build/__editable__.quakemigrate-1.1.0-cp312-cp312-linux_x86_64/quakemigrate/signal/scan.py", line 366, in locate
    self._locate_events(starttime=starttime, endtime=endtime)
  File "/home/tebw2/Software/QuakeMigrate/build/__editable__.quakemigrate-1.1.0-cp312-cp312-linux_x86_64/quakemigrate/signal/scan.py", line 505, in _locate_events
    event, _ = self.mags.calc_magnitude(event, self.lut, self.run)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/tebw2/Software/QuakeMigrate/build/__editable__.quakemigrate-1.1.0-cp312-cp312-linux_x86_64/quakemigrate/util.py", line 658, in wrapper
    result = func(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^
  File "/home/tebw2/Software/QuakeMigrate/build/__editable__.quakemigrate-1.1.0-cp312-cp312-linux_x86_64/quakemigrate/signal/local_mag/local_mag.py", line 204, in calc_magnitude
    self.mag.plot_amplitudes(
  File "/home/tebw2/Software/QuakeMigrate/build/__editable__.quakemigrate-1.1.0-cp312-cp312-linux_x86_64/quakemigrate/signal/local_mag/magnitude.py", line 466, in plot_amplitudes
    _, ax = amplitudes_summary(
            ^^^^^^^^^^^^^^^^^^^
  File "/home/tebw2/Software/QuakeMigrate/build/__editable__.quakemigrate-1.1.0-cp312-cp312-linux_x86_64/quakemigrate/plot/amplitudes.py", line 160, in amplitudes_summary
    ax, _ = label_stations(ax, rej_trids, rej_amps, rej_dists, rejected=True)
[...]

PR #181, opened 2024-09-09
FIX (plot/amplitudes.py) Explicitly use pd.Series

TomWinder:
Address #180 by explicitly using `pandas.Series` objects for input to `label_stations()` in `plot.amplitudes`.

Also fix an annoying plotting bug by removing the redundant `marker` argument (and adding the required `fmt` argument for plotting used amplitude observations).

### Relevant Issues
#180 

## Test cases

This bug was allowed to occur because it is not currently covered by a test. It is difficult to write a test other than using an example. I have locally added a trace filter to the Volcanotectonic_Iceland example such that only horizontal-component data is used (hence generating rejected amplitude measurements from the vertical channels). This previously failed with the above error, but now runs without any issues. (And the amplitudes plot is fixed!) The addition is:

```diff
+ mag_params.trace_filter = ".*H[NE]$"
```

If we retain this change it will require the benchmarks to be updated - I think on balance this is worth doing. I have not yet committed this change here.

EDIT: this update has now been implemented, and the VT_Iceland benchmark updated accordingly. All tests pass with update benchmarks.

### System details
Ubuntu 22.04 LTS
Python 3.12.2
QuakeMigrate 1.1.0

### Final checklist
- [x] All tests still pass.


hemmelig:
I agree, worth changing benchmarks in order to include this part of the code in the testing process. Go ahead with the change + benchmark update. Follow that commit up with another incrementing the version to 1.1.1.

hemmelig:
I see two arguments for leaving as-is:

1. As you say, people tend to just lift parameters straight from the examples, so keeping them as broadly appropriate as possible seems sensible
2. The benchmarks are there to test the codebase is unchanged between changes we make/changes made to our dependencies, so doesn't matter if they aren't completely 'sensible'

Will give it a quick scan and hopefully approve shortly. Do you want to try the release candidate process?

hemmelig (approved):
Changes address the described issues:

1. Fixes a bug with rejected measurements
2. Addresses a plotting bug that was causing the amplitude plots to connect scatter points with lines
3. Ensures the rejected measurements section of the magnitude computation code is covered by the examples + benchmarks testing

Benchmarks have been updated accordingly, so looks good to merge.

PR #182, opened 2024-09-12
FIX: minor local magnitudes / amplitudes bugs

TomWinder:
Minor bugfixes within the local magnitudes sub-module. The first adds handling for the case where no amplitude measurements are made. The second fixes application of the (relevant - depending on phase) filter gain correction to the measured noise amplitude, within magnitude calculation and amplitude summary plot plotting.

### Relevant Issues
N/A

## Test cases

  - BUT: note I hadn't noticed in #181 that there _is_ one event (`20140824000752340`) that does not have any "valid" amplitude observations (based on the filters) and therefore doesn't get a magnitude, and there are two events (`20140824000321580` and `20140824000512920`) which have only one amplitude(/magnitude) estimate, leading to a divide by zero error / RuntimeWarning: `quakemigrate/signal/local_mag/magnitude.py:912: RuntimeWarning: divide by zero encountered in scalar divide
  r_squared = (log_amp_variance - mod_variance) / log_amp_variance`. This might push over the edge towards adjusting the filters; I'm not sure it's good to be asking for divide by zero nonsense in the example/test. (And having a warning shown to the user).

### System details
Ubuntu 22.04 LTS / Python 3.12.4 / AMD Epyc

### Final checklist
- [x] All tests still pass.


hemmelig (commented):
My reading is that this is mostly a reorganisation and a sanity check.

Would it be worthwhile/reasonable to extend the check for 0 amplitude measurements to cover instances where only 1 is made? While it might well be a useful measure, without any uncertainty it is pretty hard to tell. As you say, having a warning show up in the example, undocumented, seems less than ideal.

hemmelig (approved):
Approved. Fine to direct at master, will need to rebase the version/v1.2.0 branch once done and handle any conflicts.

Issue #173, opened 2024-03-07
Warnings about deprecation of Node.js 16


TomWinder:
**Describe the bug**
Deprecation message on GitHub Actions workflows requesting to update to Node.js 20 actions

**To Reproduce**
See GitHub actions reports.

**Additional context**
From reading through the linked file, it looks like it should be as simple as changing:
```
      - name: Set up Python ${{ matrix.python-version }}
        uses: actions/setup-python@v3
```
to
```
      - name: Set up Python ${{ matrix.python-version }}
        uses: actions/setup-python@v4
```
etc. (for `checkout`, `upload-artifact`, `action-tmate` and `download-artifact`, at least)

.. But I'm not sure if there are any knock-ons in terms of syntax etc. that need to be updated together with this.

![Screenshot 2024-03-07 at 17 57 59](https://github.com/QuakeMigrate/QuakeMigrate/assets/40000578/2946f735-7a2c-445d-afaa-2fc3bd78c851)

hemmelig:
I guess we just create a new branch and follow what happens. If we need to add more, we can squash it all into a single commit before merging.

PR #183, opened 2024-09-13
Fix coverage reporting via CodeCov

hemmelig:
This PR will fix an issue with our coverage reports, which were failing to be properly parsed by CodeCov. The paths in the .xml report were absolute paths to the site-packages directory on the run machine, rather than relative to the package top-level (`quakemigrate`).

Add a codecov.yml file that should adjust these paths when sending the report. Also added the tests for specific parts of the package (under `tests/`) to the coverage report.

### Relevant Issues
No relevant issue.

## Test cases
This PR does not change the source code at all. The GitHub Action for running our tests was successful and the coverage report was correctly parsed by CodeCov.

### System details
macOS 12.5.1 (M1 chip)

### Final checklist
- [x] All tests still pass.

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -35,7 +35,7 @@ jobs:
           fetch-depth: 0
 
       - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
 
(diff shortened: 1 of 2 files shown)


hemmelig:
Final question is whether we should re-introduce the "Only publish the coverage report for releases".

This is the result on codecov.io:

https://app.codecov.io/github/QuakeMigrate/QuakeMigrate/commit/6dd40ba1d962e8698fc3d420aaffc72ceba041ee

TomWinder:
Looks good. Could we use this opportunity to try `setup-python@v5` to also address #173? Had a quick look at the release notes and I think it should "just work".

TomWinder (approved):
Solves the issue, as well as resolving #173.

PR #184, opened 2024-09-19
FEAT: Askja VT-DLP example

TomWinder:
Adds a new example from Askja volcano, Iceland, which will be featured in the forthcoming manuscript. Additionally adds functionality to output the marginalised coalescence maps produced during `locate()`, which may be used for plotting purposes.

Remaining possibilities not yet addressed:
~- [ ] Add a benchmark to be used in the testing. This would add coverage for trigger smoothing.~
    - to be addressed later, after benchmark storage strategy has been overhauled
~- [ ] Upgrade the VT_Iceland example get_data script to also use the obspy `MassDownloader`~
    - to be added in a future PR, where this will be rolled out to all examples.

### Relevant Issues
N/A

## Test cases

### System details
Ubuntu 22.04 / Python 3.12.5 / AMD Epyc

### Final checklist
- [x] All tests still pass.


hemmelig (commented):
Example looks great—I will run it on my machine (macOS).

Only reason not to add benchmarks would be if it takes a long time to run. Part of me is thinking we need to review how we store our examples and benchmarks. Every small change to the benchmarks will lead to bigger and bigger GitHub downloads for anyone wanting to build from source OR run the examples locally. I have some thoughts but want to flesh them out a bit.

If it all runs fine, I'll approve this PR.

hemmelig:
I think we should disable pick plotting for the tests with many events. We can leave it on for the icequake example in order to test that part of the code hasn't broken (well, to the point of the program failing, we obviously don't have any comparison of the outputs). This would massively cut down the run time.

hemmelig:
Ok, the example ran fine on my system (macOS 12.5.1, M1 chip, Python 3.12)

Let's add the benchmarks and then I think this is good to go.

TomWinder:
Hmm I think the barrier for the benchmark is probably going to be the size of the LUT file (66MB), while git will raise warnings if you to try to push anything bigger than [50MB](https://docs.github.com/en/repositories/working-with-files/managing-large-files/about-large-files-on-github#file-size-limits).

So that probably pushes me towards holding off until we've found a more sustainable solution for storing test data.

Re. pick plots: yes I often turn it off manually before running for exactly this reason. However, I was also fiddling some more with the super slow plotting I was experiencing the other week and found the backend can make a _huuuuge_ difference - even between QtAgg and Agg, with Agg still being multiple times if not an order of magnitude faster. Even had a brief think about enforcing Agg if the user selects "interactive=False" for the trigger plot, and always for locate, but it seemed like there might be a chance of unintended consequences. Would be interested to know if you experience the same thing on macOS.

So: I will turn off pick plots for VT_Iceland and for this example, and switch to `MassDownloader` for the VT_Iceland example (also gives a big speedup). And add stuff to the changelog.

hemmelig:
Can we set up a GitHub LFS specifically for this, then during the tests use the checkout action to pull in that data? Would take a bit of fiddling I expect to figure out, so perhaps not something for right now.

TomWinder:
Maaaybe. But I really hated the experience of using LFS previously - and I imagine the traffic restrictions could easily be a problem for repeatedly pulling to run the tests. 1GB per month I think. (We kept hitting this in the day before fieldwork trying to pull to ~8 laptops).

hemmelig:
Alternatively something like Zenodo and just curl the data from there? The benchmarks could be versioned to match the release they pertain to. Adds a layer of faff to properly manage and keep the Zenodo release updated, I suppose.

Or, we don't create a LUT benchmark for this example and rely on the other examples.

TomWinder:
Summarising external discussion:
 - test coverage added by the Askja example is restricted to using trigger smoothing, and this is already covered by specific unit tests
 - best to delay adding this (if we choose to) until we have revised the benchmark storage strategy -- and also perhaps the strategy for separating benchmark scripts from example scripts which are likely to be used as a starting point for users
 - amongst the discussion re. storage usage it was identified that we are also currently shipping ~ 115 MB of Rutford data with the repo, which is doubly unnecessary because (a) it's available on IRIS/EarthScope and (b) we don't need 1 hour of data to scan 5 seconds of it for events.... ; this will be addressed in a separate following PR (as well as the Iceland icequake example). The VT_Iceland get_data script will also be replaced with the MassDownloader.

So following commits will address the other residual bits that were promised:
 - turn off pick plotting for VT_Iceland and Askja examples
 - add changes to CHANGES.md

hemmelig (approved):
Good to go. I noticed a few minor formatting things with the example scripts (2 lines after imports, one line between sections of the script, 2 lines before/after function definitions), but don't want to touch these right now.

Issue #158, opened 2023-05-04
Allow multiple channels for P phase in event & pick plots


TomWinder:
**Is your feature request related to a problem? Please describe.**
Currently the event summary plot, and pick plots, are hardcoded to consider only Z-component data for P, and [N,1] & [E,2] data for S. This means if you e.g. add in the OBS hydrophone channel to the P-phase channel mapping it isn't plotted.

**Describe the solution you'd like**
Allow multiple channels to be plotted for P; use channel-mapping for both P & S to remove hard-coding.

**Additional context**
Inspired by @shicks-seismo inquiry about incorporating OBS hydrophone data for P-phase onset calculation. (Note that a quick test shows this *is* successfully being used for onset & location calculation, just not plotted anywhere for easy inspection!)

PR #185, opened 2024-09-23
FEAT: MassDownloader rollout

TomWinder:
Roll out the MassDownloader to move from bundling data for the included examples to downloading it from IRIS/EarthScope at runtime. Also move from Client -> MassDownloader for the `Volcanotectonic_Iceland` example, to take advantage of the efficiency gains it provides.

Secondarily, address #158 (and allow the `Icequake_Rutford` example to still produce plots with the IRIS-provided data, which is labelled Z->1, N->2 and E->3 for the three components) by adding the capability to determine component lists for each phase from the user-supplied `onset.channel_maps` dict.

Not yet done:
 - [ ] Update notebooks for the two Icequake examples (always forget these..)

### Relevant Issues
#158 

## Test cases

    - `Icequake_Iceland` raw cut waveforms have been updated, as detailed in the commit.

### System details
Ubuntu 22.04 / Python 3.12.5 / AMD Epyc CPU

### Final checklist
- [x] All tests still pass.
- [ ] Any new features or fixed regressions are covered by new tests.
- [ ] Significant changes have been added to `CHANGES.md`.


quakemigrate/plot/event.py
```diff
@@ -100,13 +100,31 @@ def event_summary(run, event, marginalised_coa_map, lut, xy_files=None):
     _plot_text_summary(text, lut, event)
 
     # Deal with repeating labels in waveform gather legend; combine labels for
-    # ["N", "1"], ["E", "2"]
+    # e.g. ["N", "1"], ["E", "2"]; and same for P (e.g. ["Z", "1"]) -- based on
+    # components supplied for each phase in onset.channel_maps
+    p_str, s_str_1, s_str_2 = util.get_phase_component_strings(
+        event.onset_data.channel_maps
+    )
```
hemmelig, line 107: Hmm, just wondering if there's a technical distinction between the onset channel maps and the explicit number of channels that go into computing the onsets.

Also, want to test what happens if we run with a single phase.

hemmelig (commented):
Going to test what happens when running with a single phase before approving (and there's the other parts that we want to finish before merging).

TomWinder:
I ran with a lot of combinations of channel maps (some numbers, no numbers, matching and non-matching sets) and a single phase. Not sure if I ran with either phase explicitly set to nothing in the channel maps. Not sure if I ran with only one component for S, nor with one numeric and one non-numeric (which should be allowed). Quite annoying to navigate the possibilities within the "two components for S" plotting framework.
________________________________

hemmelig:
Ok, reviewing it, I don't think it's going to cause any breaks that I can see. Plots where non-standard phases have been used might not come out complete, but I think that's really a future issue (if at all). Approving to merge.

hemmelig:
...hmm, a little preemptive. I'll look at adding more to this to minimise what we repeat.

PR #190, opened 2024-10-21
FIX: Trigger day splitting

### LINKED ISSUE #132: BUG: Reading triggered events when using locate with partial days of data
**Describe the bug**
When running Locate for a window of time with start/end times that are not an integer number of days apart _and_ that cross a dateline, the process of reading in triggered event files will fail (silently) for the final day. It will simply not read in the triggered event file for the last day, and will only locate events for Julian days before this.

For example, running Detect, Trigger, and Locate with the `starttime`/`endtime`:

```
starttime = "2018-121T22:00:00.0"
endtime = "2018-122T02:00:00.0"
```

Any events triggered between `22:00` and `00:00` will be successfully located, but any between `00:00` and `02:00` will not.

This arises in the `quakemigrate/io/triggered_events.py' file, for lines 53-66:

```
trigger_files = []
readstart = starttime
while readstart <= endtime:
    fstem = f"{run.name}_{readstart.year}_{readstart.julday:03d}"
    file = (fpath / f"{fstem}_TriggeredEvents").with_suffix(".csv")
    if file.is_file():
        trigger_files.append(file)
    else:
        logging.info(f"\n\t    Cannot find file: {fstem}")
    readstart += 86400
if len(trigger_files) == 0:
    raise util.NoTriggerFilesFound
events = pd.concat((pd.read_csv(f) for f in trigger_files),
                   ignore_index=True)
```

The initial `readstart <= endtime` test returns `False` for the final day, if the hour/minutes of the `starttime` are later than the hour/minutes of the `endtime`.

This may also be an issue when reading the availability files and plotting during Trigger - I am investigating and will update this Issue accordingly.

**To Reproduce**
Run Detect and Trigger using a small sample of data over a dateline (start and end times either side of midnight, e.g. those above). Locate will successfully register events triggered before midnight, but not after.

**Expected behavior**
All event should be detected, as triggered event files (which are stored as day files, named by the Julian day) exist.

**Desktop (please complete the following information):**
 - Operating System: macOS Big Sur 
 - Python version: 3.8
 - QuakeMigrate version: 1.0.1

**Additional context**
This was identified when running tests on seismic data from the HVO network.

TomWinder:
Make day-splitting consistent throughout trigger (and reading of triggered events within locate), and fix #132 . 

When a multi-day QM run is processed through trigger, the treatment of the sample that falls at midnight needs to be consistent (such that the rare case of a trigger at midnight doesn't result in duplicate detections on consecutive days). This was not (fully) the case - now addressed here:
 - within a (multi-day) trigger run, if a batch endtime is at midnight, it will be adjusted back by one sample (according to the sampling interval in the input `.scanmseed` data). The implicit correction (using a "less than" condition on the endtime within `_filter_events()` is then set back to being "less than or equal to", matching the behaviour for the starttime.
 - when running locate, if the specified endtime in a call to `read_triggered_events` is at midnight, the condition when filtering the resulting events DataFrame is set to "less than" rather than "less than or equal to" as it is for any other endtime.

The adjustment of the "batchend" time also carries through to the trigger summary plot, which now specifies that it runs to YYYY-MM-DDT23:59:59, rather than running to midnight on the following day (and xlims are also correct).

Note: because of the "less than" used previously within `_filter_events()`, it would not have previously occurred that an event at midnight would be triggered twice. But it leads to undocumented unintuitive behaviour with other endtimes (in my opinion). It could have occurred that an event triggered at midnight would be located twice, though the cost of this is just wasted compute. An alternative solution to what's presented here would be to just use the "less than" for the end time in both places. As long is it's clearly documented and consistent I think either is fine.

Note 2: I'm also fairly sure there is inconsistency when looping between `while readstart < endtime:` vs `while readstart <= endtime:` -- might be worth quickly reviewing this.

### Relevant Issues
#132 

## Test cases
Tailored version of the Askja_VT-DLP example used to reproduce and then confirm that #132 was addressed by bd1a4ea -- running from 2011-10-25T23:55 to 2011-10-26T01:00

### System details
Ubuntu 22.04 / Python 3.12.5 / AMD Epyc

### Final checklist
- [x] All tests still pass.


quakemigrate/io/triggered_events.py
```diff
@@ -13,6 +13,7 @@
 import logging
 
 from obspy import UTCDateTime
+from datetime import time
```
hemmelig, line 16: This is a standard library package so should be imported on line 14 (and separated from the obspy etc imports by a blank line).

hemmelig:
Great, nice catch. Benchmarks run and tests pass on my end (macOS 12.5.1, M1 chip).

I think just move the import and update the CHANGES.md file and this is fine to move.

TomWinder:
Quick review of `< endtime` vs `<= endtime` as promised....

```
grep "< endtime" */*py
io/data.py:        while loadstart < endtime:
io/scanmseed.py:    if stats.endtime < endtime - stats.delta:
io/triggered_events.py:                (events["CoaTime"] >= starttime) & (events["CoaTime"] < endtime)
signal/trigger.py:        while batchstart < endtime:
```

```
grep "<= endtime" */*py
io/availability.py:    while readstart <= endtime:
io/triggered_events.py:        while readstart <= endtime:
io/triggered_events.py:                (events["CoaTime"] >= starttime) & (events["CoaTime"] <= endtime)
signal/trigger.py:            batchend = next_day if next_day <= endtime else endtime
signal/trigger.py:            (events["CoaTime"] >= starttime) & (events["CoaTime"] <= endtime), :
```

Of these I think the `io/data.py:        while loadstart < endtime:` should be changed to `<=` in case that final sample over the dayline is actually required for the padding etc.; I double-checked that obspy will read the single sample. And I've tweaked the looping so that the `UTCDateTime(<time>.date)` is only done once at the start of the loop, as elsewhere. In trigger this is done within each stage of the loop, but that's necessary because the time is important for `batchstart`, not just the date.

My final thought on this was whether it would technically be more robust to use `timedelta(days=1)` to skip forwards 24h, but some quick reading about datetime's handling of leap seconds (https://github.com/python/cpython/issues/67762) suggests this isn't likely to cause a problem any time soon. Made me reflect again about how this is crudely ignored in seismology, and whether that actually does cause some problems (can easily imagine teleseisms having phase arrivals spread both before and after a leap second, for example, but I digress...)

hemmelig (approved):
Happy to approve.

TomWinder:
Ha. But also - wait - hold off this until #168 is actually merged into v1.2.0, else this will cycle again (didn't think about that before handling the conflict here, of course...)

hemmelig (approved):

PR #194, opened 2024-12-14
BUGFIX: trigger plot failure when no events

hemmelig:
Fix an issue with the trigger plotting function when no events were available to be plotted. Now, instead of `None`, an empty Pandas DataFrame is passed in to the plotting function, but the if statement still expected None when no events to be plotted.

### Relevant Issues
None

## Test cases
Ran the trigger stage for a day of data for which the trigger plotting issue had arisen (no events triggered). Ran successfully.

### System details

### Final checklist
- [x] All tests still pass.

### DIFF
--- a/quakemigrate/plot/trigger.py
+++ b/quakemigrate/plot/trigger.py
@@ -148,12 +148,12 @@ def trigger_summary(
         _plot_event_scatter(fig, discarded_events, discarded=True)
 
     # --- Plot event scatter on LUT and windows on coalescence traces ---
-    if events is not None:
+    ax_i = 1 if normalise_coalescence else 0
+    if not events.empty:
         _plot_event_windows(fig.axes[:2], events, marginal_window)
         _plot_event_scatter(fig, events)
 
         # Add trigger threshold to the correct coalescence trace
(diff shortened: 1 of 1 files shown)


TomWinder (approved):
Good spot. I guess just missed during 4540657 . Slightly difficult to add a test that covers no events being triggered...

I moved all the lines related to the coalescence trace plotting. I guess there is an argument for leaving it off (in case it zooms out too far and doesn't allow you to see where to put it), but ...

hemmelig:
I'm going to merge this tomorrow, as is, unless you've any objections—i.e., think we should test build and do a proper release. Otherwise, we can just get this to master and likely it would end up bundled into 1.2.0. Seems a bit much to do an entire release for a single bug fix. I think the right balance would be every other week or so, any new bug fixes are formally put together as a release, rather than doing a release every time we merge something.

hemmelig:
Actually, yeah, think I'm going to proceed as described. Python itself has a bug fix release cadence of 2 months, so probably just follow that idea, with the next bug fix release in mid-January or so.

TomWinder:
No I agree with just merging, and trying to find some kind of routine for releases.

PR #195, opened 2024-12-16
BUGFIX: Make timestep handling explicit and uniform in detect/locate

TomWinder:
This commit hopefully ties up the thread that started being pulled in PR #190. The process of splitting up processing into timesteps (and hence the follow-on consequences for splitting a longer duration run into days) was quite unclear, and had a few bugs in it. Here this is hopefully made much more explicit and easy to understand. A bug where a detect run that runs over a day boundary by only one timestep would result in that final timestep not being written to scanmseed is also addressed.

In more detail:
 - within signal/scan.py, explicitly calculate the timestep end / endtime of a run, and warn user if the two are incompatible. If endtime is exactly midnight, adjust it to be one sample earlier, to match the (now) package day-splitting strategy. If endtime doesn't encompass an integer number of timesteps, extend the endtime accordingly.
 - in doing so, calcation of n_steps is moved from _continuous_compute() up to detect(), so the internal function now processes a specified number of timesteps.
 - adjust w_end specification in _continuous_compute(), to address a fenceposts bug (previous one extra sample of data was being read at the end of the window, leading to the fudged trimming applied in io/scanmseed.py to make things split as intended at timestep boundaries.
 - with the above changes, the aforementioned fudge has been removed from io/scanmseed.py. In addition, where a timestep crosses a day boundary but not all data from the timestep is written, self.written is set back to False, so it will be caught by the 'if not self.written' designed to mop up any residual data.
 - benchmarks are unaffected other than a single sample of the "COA" trace in the Iceland_Icequake scanmseed, which differs by 1e-5; this is presumably an edge effect from the w_end change, which results in a slightly different filter specification etc.
 - core/lib.py: add some debug statements to follow map_4d / onset input shapes
 - io/data.py: add some debug statements to check data length (related to tracing the effects of the w_end adjustment).

### Relevant Issues
N/A

## Test cases
 - Run detect over a dayline by only a single timestep --> scanmseed data after dayline now written to file successfully.

### System details
Ubuntu 22.04 / AMD Epyc / Python 3.12.5

### Final checklist
- [x] All tests still pass.
- [-] First time contributors have added your name to `CONTRIBUTORS.md`.


quakemigrate/core/lib.py
```diff
@@ -93,8 +95,11 @@ def migrate(
 
     *grid_dimensions, n_luts = traveltimes.shape
     n_onsets, t_samples = onsets.shape
+    logging.debug((n_onsets, t_samples))
     n_samples = t_samples - first_idx - last_idx
+    logging.debug(n_samples)
     map4d = np.zeros(tuple(grid_dimensions) + (n_samples,), dtype=np.double)
+    logging.debug(map4d.shape)
```
hemmelig, line 102: I think this is probably fine, but we might consider adding some information making clear what these numbers are when they are printed in the debug log. I can think of some scenarios where this would be useful, so think would just be more complete to have:

```suggestion
    logging.debug(f"(n_onsets, t_samples) : ({n_onsets}, {t_samples})")
    n_samples = t_samples - first_idx - last_idx
    logging.debug(f"n_samples : {n_samples}")
    map4d = np.zeros(tuple(grid_dimensions) + (n_samples,), dtype=np.double)
    logging.debug(f"map4d shape : {map4d.shape}")
```

quakemigrate/io/data.py
```diff
@@ -608,6 +608,8 @@ def check_availability(
                     if len(st_id) > 1:
                         continue
                     elif st_id[0].stats.npts < n_samples:
+                        logging.debug("Trace has too few samples.")
+                        logging.debug((n_samples, st_id[0].stats.npts))
```
hemmelig, line 612: Same here, just a little additional clarity:

```suggestion
                        logging.debug(f"(n_samples, trace_npts) : ({n_samples}, {st_id[0].stats.npts})")
```

quakemigrate/signal/scan.py
```diff
@@ -307,16 +309,32 @@ def detect(self, starttime, endtime):
...
+
+        # Number of steps to break run duration into
+        n_steps = int(np.ceil((endtime - starttime) / self.timestep))
+
+        # Check if chosen start & endtimes and timestep are compatible
+        calc_endtime = starttime + n_steps * self.timestep - 1 / self.scan_rate
+        if calc_endtime - endtime > 1 / self.scan_rate:
+            logging.info(
+                f"Warning: chosen run duration {endtime - starttime} s is not divisible "
+                f"by the specified timestep {self.timestep} s.Detect will instead "
```
hemmelig, line 325: Missing space between "s." and "Detect".

quakemigrate/signal/scan.py
```diff
@@ -417,6 +436,7 @@ def _continuous_compute(self, starttime, endtime):
             try:
                 data = self.archive.read_waveform_data(w_beg, w_end)
                 time, max_coa, max_coa_n, coord, onset_data = self._compute(data)
+                logging.debug(max_coa.shape)
```
hemmelig, line 439:
```suggestion
              logging.debug(f"1-D con shape : {max_coa.shape}")
```

hemmelig (commented):
I have added a few specific comments that I think should be addressed, but they're pretty minor. Once done, I'll run everything locally on my end and approve.

TomWinder:
Thanks for checking through. Agree with all the suggestions to add context for this stream of integers otherise being printed to the output...

Could you also just go over my logic checking the timesteps & fenceposts (particularly handling ends of days and batches overlapping daylines)? I have stared at it a while and think it holds water, but slightly difficult to convince myself I've thought through all the edge cases.

hemmelig:
Maybe because it's being merged into v1.2.0, and so is pulling off a branch that predates the v1.1.2 that is currently on master, but it's installing as v1.1.1. I think fine, as we will address this later when we bump to v1.2.0 in the pyproject.toml file.

hemmelig:
Yep, looked over the tilmestep changes, see how it solves the midnight on last day issue (good). Can't see any other issues, logic looks sound to me.

hemmelig (approved):
Approved to merge.

PR #196, opened 2024-12-16
FEAT: Add option to only plot active stations in event summary

TomWinder:
Previously all stations in the LUT were always plotted in the locate event summary. This commit adds the option to plot only stations that have contributed onset functions that are used for the location. In doing so that station is omitted from both the map/cross-sections and the waveform gather.

This can be helpful when running over a long time period with a large and temporally variable network, where otherwise the waveform gather can be dominated by empty rows.

### Relevant Issues
N/A

## Test cases
Specified `plot_all_stns = False` in the Icequake_Iceland example (as here station SKG09 is not present for the time window we are running). All worked as expected.

### System details
Ubuntu 22.04 / AMD Epyc / Python 3.12.5

### Final checklist
- [x] All tests still pass.
- [-] First time contributors have added your name to `CONTRIBUTORS.md`.


hemmelig (commented):
Neat. I'll run locally, but don't foresee any issues. Once done, will approve. Pretty sure we do the distance ordering during the waveform gather plotting, so shouldn't matter we're passing an alphabetical list of stations, but that's the only thing I can think of to check.

hemmelig:
Yep, just checked the code and we do indeed just do the ordering during the plotting stage, so see no issues.

hemmelig (commented):
Tested locally and all tests pass. Toggled the `plot_all_stns` to `False` for the Iceland ice quake example, as one of the stations doesn't contribute, and had no issues.

Final query is whether we should just use `plot_all_stations` rather than `plot_all_stns`, for the complete avoidance of doubt? A few extra characters, but no harm in being precise. Let me know, otherwise happy to approve.

TomWinder:
Yep I also considered this when re-reading it before pushing. I think the reason I had it as this (and definitely the reason I left it) was because we use `plot_all_stns` for the equivalent option for the trigger summary plot. Lmk what you think - I'm on the fence. Would probably want to change both if making this `plot_all_stations`

I'll push the `CHANGES.md` to both then ready to merge I think, pending decision on this.

hemmelig (approved):
Approving post discussion on variable naming. A full review at some future point seems worthwhile, but will involve some managed deprecation so want to do all at the same time if at all possible.

PR #200, opened 2025-01-21
:bookmark_tabs: bump version to v1.2.0 EOM

hemmelig:
Bump version number in `pyproject.toml` file to 1.2.0 (since all `version/` branches are protected).


TomWinder:
Do the formalised v1.2.0 changes not also want to be finalised here? (i.e. add v1.2.0 header, check if anything was missed)

PR #203, opened 2025-01-24
Resolve merge conflict v1.2.0

hemmelig:
`master` and `version/` branches are protected, so to fix any merge conflicts this has to be done (as we didn't rebase the version branch when the bug fix went out to `master`). Might be worth thinking this process over for future in instances where we push a bug fix to master while working on a new version.

## Test cases
Ran test example and benchmark test.

### System details
macOS 12.5.1 M1 chip

### Final checklist
- [x] All tests still pass.


TomWinder (approved):
I guess we should have a system to keep the new version branch up to speed with any bugfixes being applied directly down to master. But looks good.

PR #205, opened 2025-04-09
:bug: update PyPI publish action

hemmelig:
Update PyPI publish action from v1.9.0 to v1.12.4, which should handle issues with package metadata that was causing the publish workflow to fail.

### Relevant Issues
None

## Test cases
No changes to codebase, so tests still pass.

### System details
Not tested, as this is run as a GitHub Action.

### Final checklist
- [x] All tests still pass.

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -86,7 +86,7 @@ jobs:
           name: artifact
           path: dist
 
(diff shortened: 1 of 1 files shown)


hemmelig:
Just recording this here. The source of the issue, and it's fixed, was discussed here:

https://github.com/pypa/gh-action-pypi-publish/issues/310

TomWinder (approved):

PR #206, opened 2025-05-16
bugfix: updated get_phase_component_strings() to plot when no S picks

acad3:
I am trying to use QuakeMigrate to identify ice quakes but my array consists only of vertical component nodes. Every other aspect of the code works when setting the phase to only P, except for the function get_phase_component_strings(), causing the code to error when running the relative relocation plotting.  Have modified to include an if statement, so only gets and returns the S strings if S is a valid channel. 

### Relevant Issues
Link any open Issues here.

## Test cases

- Rutford ice stream example runs as written even with new changes, but if you set the phase list to only be P in all instances, no longer errors at scan and can run through the whole example.

### System details
Ubuntu 20.04.6 LTS

### Final checklist
- [ ] Correct base branch selected?
- [ ] All tests still pass.
- [ ] Any new features or fixed regressions are covered by new tests.
- [ ] Significant changes have been added to `CHANGES.md`.
- [ ] First time contributors have added your name to `CONTRIBUTORS.md`.  --> didn't feel this was really necessary for an if statement

### DIFF
--- a/quakemigrate/util.py
+++ b/quakemigrate/util.py
@@ -683,6 +683,9 @@ def get_phase_component_strings(channel_maps):
     that flexible) plotting setup. If the use of this function is expanded beyond that
     of a plotting utility, this might need to be revised.
 
+    Ammended to facilitate plotting when no S picks are made, e.g. in case of having
+    vertical components only, through if statement.
+
     Currently all P-phase components are retained; this means if more than one
     component is specified for the P onset, the waveforms will be overlain.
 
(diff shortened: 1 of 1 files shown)


hemmelig:
Hey! Very cool that you're using QM—where's your dataset from?

Yeah, very possible to run in just P-phase mode, but definitely some assumptions bundled into the plotting functions about the number of components present, which components they are, etc. If you've got things working for you with this adjustment, I recommend using it for now. I'll have a think through how we are handling things right now, as it definitely feels like a slightly deeper problem worth addressing.

Issue #204, opened 2025-03-17
Failing to plot trigger summary plots with 'trig.plot_all_stns = False' after iteritems is removed from pandas 2.0


esmesouthern:
**Describe the bug**
Trigger summary plots fail to plot after iteritems is removed from pandas 2.0. 

**To Reproduce**
Running trigger with latest version of pandas and with 'trig.plot_all_stns = False' fails to produce summary plots and produces the following error:

```
Traceback (most recent call last):
  File "/home/eos24/quakemigrate/RUN_SCRIPTS/trigger_svartsengi_GAUSS_v_1_2_0_test.py", line 79, in <module>
    trig.trigger(start, end, region=region)#, interactive_plot=False)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/eos24/miniconda3/envs/qm_1_2_0_2/lib/python3.12/site-packages/quakemigrate/signal/trigger.py", line 305, in trigger
    self._trigger_batch(batchstart, batchend, region, interactive_plot)
  File "/home/eos24/miniconda3/envs/qm_1_2_0_2/lib/python3.12/site-packages/quakemigrate/signal/trigger.py", line 374, in _trigger_batch
    trigger_summary(
  File "/home/eos24/miniconda3/envs/qm_1_2_0_2/lib/python3.12/site-packages/quakemigrate/util.py", line 658, in wrapper
    result = func(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^
  File "/home/eos24/miniconda3/envs/qm_1_2_0_2/lib/python3.12/site-packages/quakemigrate/plot/trigger.py", line 132, in trigger_summary
    for col, ava in availability.iteritems():
                    ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/eos24/miniconda3/envs/qm_1_2_0_2/lib/python3.12/site-packages/pandas/core/generic.py", line 6299, in __getattr__
    return object.__getattribute__(self, name)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'DataFrame' object has no attribute 'iteritems'. Did you mean: 'isetitem'?
```

**Desktop (please complete the following information):**
 - Operating System: Linux
 - Python version: 3.12
 - QuakeMigrate version: 1.2.0
 
 A workaround:
 import pandas as pd
[...]

PR #209, opened 2025-06-21
:bug: Remove use of removed `pandas.DataFrame.iteritems()`

TomWinder:
As described in #204, `DataFrame.iteritems()` was removed in pandas 2.0, resulting in a failure within trigger plotting. This commit implements the discussed (and recommended) alternative `DataFrame.items()`. Also verified that there were no further instances of this function being used within the package.

### Relevant Issues
#204 

## Test cases
Tested the `Icequake_Iceland` example with `trig.plot_all_stns = False` (as described by @hemmelig in #204) to verify that this change resolved the failure - and resulted in a trigger summary plot with station "SKG09" excluded.

    - Note: I am currently receiving a LUT-related test failure, but this was seen both before and after this change, and is not relevant.

### System details
Apple M4 Pro / macOS 15.5 / Python 3.12.11

### Final checklist
- [x] All tests still pass.

### DIFF
--- a/quakemigrate/plot/trigger.py
+++ b/quakemigrate/plot/trigger.py
@@ -129,7 +129,7 @@ def trigger_summary(
     if availability is not None:
         station_list = []
         if not plot_all_stns:
-            for col, ava in availability.iteritems():
+            for col, ava in availability.items():
                 if np.any(ava == 1):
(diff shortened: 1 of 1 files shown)


hemmelig (approved):
Approved.

PR #231, opened 2026-01-08
FEAT: Add the option to ignore network & loc codes in waveform archives

TomWinder:
Add the option to ignore network codes and location codes in the waveform data archive. This may be useful when a given station (or set of stations) have been assigned different temporary network codes at different times during the study period; i.e. not reflecting any physical change to the station infrastructure, but which would otherwise prevent waveforms from being merged due to their incompatible SEED ID's. This is also reflected in the output cut waveforms.

This is achieved through the addition of three new attributes to the `Archive` class, and corresponding kwargs to the funciton `quakemigrate.io.core.read_inventory` so that the wiping/substitution of codes is mirrored in the response info:
 - `ignore_network_code`
 - `dummy_network_code`
 - `ignore_location_code`

The dummy network code is checked to ensure it is a two-character string. If not specified it is set to 'XX'.

### Relevant Issues
N/A

## Test cases
- [x] Added the `ignore_network_code=True` kwarg to the `Iceland_Icequake` example; runs successfully but output cut waveforms now have network code 'XX'.
- [x] Attempted to set an invalid network code ('XXX') and this was rejected.
- [x] Set `ignore_location_code=True` after manually editing location code for the input mSEED; runs successfully but output cut waveforms now have location code "" (empty string).

### System details

### Final checklist
- [x] All tests still pass.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,8 @@
+- quakemigrate.io.data
+  * Add the option to ignore network codes and location codes in the waveform data archive. This may be useful when a given station (or set of stations) have been assigned different temporary network codes at different times during the study period; i.e. not reflecting any physical change to the station infrastructure, but which would otherwise prevent waveforms from being merged due to their incompatible SEED ID's. This is also reflected in the output cut waveforms.
+- quakemigrate.io.core
+  * Add the option to ignore network codes and location codes in the instrument response inventory; see corresponding change in quakemigrate.io.data for further details.
+
 1.2.2
 =====
 - GH Actions
--- a/quakemigrate/io/core.py
+++ b/quakemigrate/io/core.py
@@ -3,7 +3,7 @@
 Module to handle input/output for QuakeMigrate.
 
 :copyright:
-    2020–2023, QuakeMigrate developers.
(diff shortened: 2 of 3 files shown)


hemmelig (approved):
Definitely a useful shorthand to avoid faff, I like it.

I don't think it makes a difference, insofar as having the `ignore_X` is a consistent interface, but I suppose you could either just set the dummy network code implicitly, drop the explicit argument, or drop the `ignore_network_code` argument and just default the dummy network code to `None` and do a `if dummy_network_code not None:`.

I will approve anyway, as the above thoughts change nothing.

hemmelig:
Actually, re-requested a review to try cancel the approval (may or may not affect the ability to merge) as I realise another aspect of the PR is where should this go.

I think as this is a very minor improvement that has no impact on pre-existing implementation it could just go straight to master and either be bundled in as part of a patch (if we accrue a few more bug fixes by end of February, say) or a minor version otherwise?

TomWinder:
Yes agree with putting this straight in master.

I see the point of adding only the `dummy_network_code` kwarg, but I think the only possible benefit is a slight faff-saving, so considering this is already included I think it's fine to stay as-is.

Issue #239, opened 2026-03-25
Internal usage of a Station dataclass


hemmelig:
**Is your feature request related to a problem? Please describe.**
Currently, we are using pandas to pass around our station information. We do not have the option to use network or location codes as a way of distinguishing stations, and rely only on station codes. This would break in instances where two stations in different networks share names, which might be an issue when requesting data from some external data source.

**Describe the solution you'd like**
I think internally we should use our own lightweight representation of a Station, which includes fields for FDSN network codes, locations, channels. It is a straightforward switch, but does touch a few parts of the code and there are some implications we may wish to consider (e.g., do we continue to use just station codes for plotting purposes, or switch to partial SEED IDs (`NET.STA.LOC`)?

Other benefits/considerations:
- Expose the elevation and depth as distinct attributes, making the reference frame usage throughout the code explicit
- Can specify the channel codes to use at this stage on a per-station level, which would be used at the data read stage
- We could perhaps infer the recent additions of `ignore_network_code`, `ignore_location_code` from this dataclass? For example:

```python
if not any(station.network for station in stations):
    # All network codes are None, infer ignore_network_code = True

if not any(station.location for station in stations):
    # All location codes are None, infer ignore_location_code = True
```

and, since it does not really matter what dummy network code is used, we could drop it as a configurable parameter?

```python
@dataclass
class Station:
    station: str

    longitude: float
    latitude: float
    elevation: float
    depth: float

    network: str | None = None
    location: str | None = None
    channels: str | None = None

    def __str__(self) -> str:
        """Partial SEED ID representation of the station."""
        return ".".join(
            self.network if self.network is not None else "",
            self.station,
            self.location if self.location is not None else "",
        )
```

I/O function:

```python
def read_stations(station_file: str, **kwargs: dict) -> list[Station]:
    """
    Reads station information from file.

    Parameters
    ----------
    station_file:
        Path to station file.
        File format (header line is REQUIRED, case sensitive, any order):
            Latitude, Longitude, Elevation (units matching LUT grid projection;
            either metres or kilometres; positive upwards), Name
    kwargs:
        Passthrough for `pandas.read_csv` kwargs.

    Returns
    -------
    stations:
        List of Station objects.

    Raises
    ------
    StationFileHeaderException
        Raised if the input file is missing required entries in the header.

    """

    stations = pd.read_csv(station_file, **kwargs)

    try:
        stations = sorted(
            [
                Station(
                    network=station.get("Network"),
[...]

Issue #234, opened 2026-01-28
Refactoring waveform data i/o to accommodate non-local filesystem data


hemmelig:
**Is your feature request related to a problem? Please describe.**
It is more a limitation than a problem. Currently, QuakeMigrate is designed to read waveform data from an archive that is stored on the local filesystem. However, there are some circumstances where data access via an ObsPy-like Client may be of use. This is reasonably straightforward to achieve, with the current `Archive.read_waveform_data` being remapped to the `Client.get_waveforms` style used in ObsPy clients.

**Describe the solution you'd like**
At this stage, a discussion on what a redesigned data i/o system could look like. I have some proposals/examples I will include in the comments below. Some considerations/topics to discuss:

- Pre-downloading
- Operation in real-time
- Asynchronous data access (essentially, the data for the _next_ timestep is loaded and pre-processed on a dedicated while the compute engine is running the migration and stacking on the current timestep).
- Currently the path structures/archive format stuff on the `Archive` class are providing 2 different but potentially confusing ways of specifying a local archive filepath structure. I have some thoughts on how to address
- Can all of this be done without creating any breaking changes?

**Describe alternatives you've considered**
There is already a filesystem Client available through ObsPy, designed with the SDS (SeisComp3) archive structure in mind.

**Additional context**
While this would facilitate reading data from a remote archive, two things should be kept in mind:

1. The remote data server might have some degree of rate-limiting of access in place. Repeated accesses is not really an appropriate means of running QM in a sustainable fashion. Where possible (e.g., for non-real-time studies) the waveform data should be downloaded and archived locally before running any analysis.
2. Data i/o over the internet is going to be slower than from a local filesystem. It may reach the point where the whole process becomes i/o bound (meaning it spends more time reading data than on the analysis).

TomWinder:
My very initial thoughts on this are:
 
1. It seems very unlikely to me that truly remote waveform access will be a realistic use-case in the near future, because of the bandwidth required to keep up with the migration compute time. That's not to say it might not be useful in some scenarios, but I wonder if this is sufficiently niche that it doesn't justify a large overhaul if this is the only benefit that would necessitate that.
[...]

PR #243, opened 2026-04-07
Add waveform clients and internal Station model

hemmelig:

1. Add the concept of a waveform client to QuakeMigrate. This has been added as a new module, leaving the old Archive structure in place for now. The intent here is to normalise the interface to the Client.read_waveform_data method we have been using.

Update all example scripts to use this new interface. I have run through all of the examples and not encountered any issues.

I also tested the Iceland icequake example using the ObsPy FDSN client backend to acquire data on the fly. This is not a recommended means of operation (each timestep spent approximately 3.5 seconds on waveform io compared to the 0.5 seconds when accessing the files locally). In addition to being slower than bulk requesting the data, it would put unnecessary strain on the Data Management Center servers.

2. Define a new Station model that is used as the internal representation of a station, including location information and components of the FDSN naming scheme. This replaces the use of pandas DataFrames internally.

Currently these are collected together in an ordered (alphabetically by station name) list, but I could be convinced of the benefit of a dedicated dataclass that collects them together, provides clean iterator behaviour, and maybe gives cleaner access by key (station.id). This is not a hard need.

Breaks tests, technically, by moving to the use of station.id (NET.STA.LOC) over just station codes. However, with some remapping at test time I was able to validate that the outputs of the tests were identical in all other respects.

3. Deprecate the method of setting the Archive path structure by passing a key, instead enforcing that a formattable string is passed when instantiating a new Archive object. The motivation here was to remove any ambiguity about how the path structure should be set. The standard formats have been migrated into a dictionary that the user can use to lookup the required path format string. The change has been properly deprecated, so no existing scripts should fail. Also added a utility for validating that the formattable string did not contain any invalid placeholders.

This also adds the ability to specify the network and location codes, which has been a long-standing issue - two stations in different networks might have the same station code, which would lead to failures. The rest of the changes required for this happen elsewhere, so defaults of "*" have been used for these fields at this time.
### Relevant Issues
#96 
#234 
#239 

## Test cases

### System details
Linux 5.14.0 (kernel), Rocky 9 (OS)
macOS 12.5.4

### Final checklist
- [x] All tests still pass.


hemmelig:
I rebased this whole thing against the feature/2d-plot-plugin branch (which in turn has been rebased down the line), so should all be properly merged. Was all less easy than indicated, due to path changes making it look like refactored files were deleted/made new....anyway, should all have been caught.

